PR #6, opened 2025-01-02
AcadosFileManager configured for temp directories

JasperHoffmann:
Added the cleanup of acados files that were created into temporary folders by the AcadosFileManager.

For now adding the build and generate options is not possible, as the AcadosOcpBatchSolver does not allow those options. The possibility could be added for a later point, if the AcadosOcpBatchSolver would be updated!

A future feature would be to reduce the number of solvers created in the MPC class to only the BatchSolver.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@ TODOs.md
 exp
 wandb
 
+# for some dotfilemanagers ;)
+.envrc
 
 # acados stuff
 c_generated_code*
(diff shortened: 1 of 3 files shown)


Confectio (approved):
Looks good to me :)

sandmaennchen:
`build` and `generate` flags coming soon ..
https://github.com/acados/acados/pull/1404

PR #7, opened 2025-01-07
Upgrade and test iterate handling

Confectio:
Summary:
- Updated MPCState interface
- Implemented default_init_state_fn that can be set at initialization of the MPC and will be used for initializing the iterate of Acados if no MPCState is given and it will be used for backup in case the solve with given MPCState fails.
- Some refactoring
- More tests
- New acados version that fixes a bug
- Naming the individual solvers differently to keep the shared libraries from overwriting themselves


Confectio:
I have no idea why these tests fail, locally all tests pass. But this is something for tomorrow-me. Please feel free to review the code nonetheless already :) .

sandmaennchen:
> I have no idea why these tests fail, locally all tests pass. But this is something for tomorrow-me. Please feel free to review the code nonetheless already :) .

I suspect the file manager ..

JasperHoffmann:
The problem was that the git hash and runner id to download the acados artifacts was not aligned with the acados version in submodule! As suggested by @sandmaennchen  next time we should update the acados version in its own pull request.

leap_c/mpc.py
```diff
@@ -126,30 +120,70 @@ class MPCOutput(NamedTuple):
...
+                        zip(
+                            mpc_parameter.p_stagewise,
+                            mpc_parameter.p_stagewise_sparse_idx,
+                        )
+                    ):
+                        ocp_solver.set_params_sparse(stage, p, idx)
+                else:
+                    for stage, p in enumerate(mpc_parameter.p_stagewise):
+                        ocp_solver.set(stage, "p", p)
+    elif isinstance(ocp_solver, AcadosOcpBatchSolver):
```
JasperHoffmann, line 148: This could be a little bit slow. For now it should be fine, but maybe Acados should provide a batched version for this. Probably we should benchmark the time again before tackling such stuff :)

leap_c/mpc.py
```diff
@@ -62,7 +61,8 @@ def get_sample(self, i: int) -> "MPCParameter":
         )
 
 
-MPCState = AcadosOcpFlattenedIterate | AcadosOcpFlattenedBatchIterate
+MPCSingleState = AcadosOcpIterate | AcadosOcpFlattenedIterate
```
JasperHoffmann, line 64: For next time this could get its own pull request. Probably this will get deleted again if we only support batch solvers :)

leap_c/mpc.py
```diff
@@ -526,25 +529,27 @@ def policy(
...
         self,
         mpc_input: MPCInput,
-        mpc_state: (
-            AcadosOcpIterate
-            | AcadosOcpFlattenedIterate
-            | AcadosOcpFlattenedBatchIterate
-            | None
-        ) = None,
+        mpc_state: MPCSingleState | MPCBatchedState | None = None,
+        backup_fn: Callable[[], MPCSingleState]
```
JasperHoffmann, line 533: I think the backup_fn does not belong into __call__ but into __init__. The normal user should not see backup_fn ;)

JasperHoffmann (commented):
I think the pull request is good, but would move the backup_fn to __init__

leap_c/mpc.py
Confectio: Yeah, I agree, but I thought for now this distinction in single state and batched state keeps everything clean :) . Also I am not sure how alive the idea of supporting other solvers still is (which might not be batched) and in that case something similar to this might make sense.

leap_c/mpc.py
Confectio: 👍 yes, there is already a ticket on the board for timing

leap_c/mpc.py
Confectio: Hmm, originally it was part of init instead of call, but I moved it here, because I thought one would rather have a backup function for each specific call instead of a "general" backup function. In my view,
having the backup function in the call goes along better with our "stateless" approach here (since a backup function passed in the init and remembered, is also a little bit like a "state" of the MPC). In particular, it is also more explicit than passing the backup function in the init (you don't have to remember that there is a backup function somewhere that influences the success of this call and which you potentially want to change somewhere, its already in the signature of the call). As a smaller bonus, this also lets us keep the signature of the backup function itself very simple, while still not needing to "update" the backup function member extra every the time something changes (e.g., a likely scenario is that the backup function changes depending on the state or parameters).
Finally, I would argue that using backup_fn in call instead of init makes it not necessarily more visible for a normal user, since we provide functions like "self.policy" if easy application is desired. On the other hand, the backup function  seems as a potentially very useful and central feature for robust convergence, so maybe we should even desire for it to be more visible (and I mean it is still a keyword argument that can just safely be ignored if the user does not want to read what it does :D ).

Confectio: Yes, I think I see now what you mean with the code coupling. It would indeed be very nice if the backup function would be set initially for the MPC and no module "outside" of the MPC would have to touch it every again.

What do you think about combining both approaches?
When defining a concrete MPC class (e.g., the LinearSystem), then also a default_backup_fn and default_initial_state can be supplied to the super().__init__ (which is th MPC's init) which is how I assume you have imagined it (?). But we also keep the keyword argument in the call and IF we give a different backup function, then this backup function will be used, else the defined default (which can also be None, in case one does not want to try to recompute). This is basically how we use the MPCParameters, too (so we get bonus points for consistency :) ).
I would then say we pass the exact same input from __call__ to the default_backup_fn and default_initial_state_fn (so they can use, e.g., the current state in their computations).

About the initial_state_fn: I think a default initial state function (I hope we mean the same thing) makes a lot of sense, especially since right now we are just using the default initial state "all-0s". I just forgot to implement it :'D . 

Concerning the usage from outside: I am not sure what you mean here. I thought people would build their own scripts where they can then decide themselves how to handle things depending on what they want.

PR #8, opened 2025-01-09
Extend replay buffer

Confectio:
Summary:
    - Made ReplayBuffer work for more general data (not only transition tuples of shape s, a, r, s', d like previously).
    - Added collating of Acados Iterates.
    - Added recursive .to to move nested tensors to the correct device (and convert it to the "correct" type). In particular, this
    assumes that people want all tensors to be of the same type given in the initialization. I am not sure if we want to 
    leave it like this for now or what a good standard would be 
    (but also the function can be easily overwritten to take special cases into account if someone would wish for it).
    - Now the ReplayBuffer may return tensors of shape (batch_size, ) again instead of (batch_size, 1) (e.g., the reward or the done 
    tensors are such cases). The relevant code where this has caused a problem (SAC target calculation) now catches it itself.
    - Added testing for the ReplayBuffer.

### DIFF
--- a/leap_c/collate.py
+++ b/leap_c/collate.py
@@ -1,0 +1,97 @@
+from typing import Any
+
+import numpy as np
+import torch
+from acados_template.acados_ocp_iterate import (
+    AcadosOcpFlattenedBatchIterate,
+    AcadosOcpFlattenedIterate,
+    AcadosOcpIterate,
+)
+from leap_c.mpc import MPCParameter
+from torch.utils._pytree import tree_map_only
+from torch.utils.data._utils.collate import default_collate_fn_map
+
+
+def safe_collate_possible_nones(
+    field_data: list[None] | list[np.ndarray],
+) -> None | np.ndarray:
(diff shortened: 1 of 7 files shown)


JasperHoffmann (changes requested):
Looks amazing, could we maybe move the collate fn outside of the Buffer. I could use it as well for IL datasets :) Apart from device and dtype, it does not depend on self?

And just an understanding question, is this the new thing that you already upload the tensors on the correct device, with the collate_fn?

Confectio:
I decided that it is the easiest solution (and also prooobably most performant in many cases) to have one function that really only collates and one more function that is basically a recursive "tensor.to" that is applied afterwards. The custom_collate_fn_map actually does not depend on self at all anymore and the recursive tensor.to is only dependend on the dtype and the device as you said.
So I think we can move both, the custom collate_fn_map and the recursive tensor.to to a utils file if you want, the only reason it still is a part of the replay buffer is that the replay buffer was the only part where this was used.

PR #9, opened 2025-01-10
Fix potential u0 bug

Confectio:
Previously, it was possible to give an u0 in the mpc_input that is outside of lbu and ubu as defined in the ocp without anything complaining. If not desired (which is most likely), this is a potentially sneaky bug. This PR is for catching this bug. If not raising an exception in this case is desired, it can easily be turned off by using a new flag.


dirkpr:
Not sure if we want to handle this inside the MPC class or if it is better to do it one level higher up in the algorithm

PR #10, opened 2025-01-16
Adding new task description

JasperHoffmann:
Added the new abstract task class, to be used to declare tasks that can be learned by different algorithms.

### DIFF
--- a/leap_c/task.py
+++ b/leap_c/task.py
@@ -2,43 +2,64 @@ from abc import ABC, abstractmethod
 from typing import Any, Optional
 
 import gymnasium as gym
-import torch
+import numpy as np
 
-from leap_c.mpc import MPC, MPCBatchedState, MPCInput
+from leap_c.mpc import MPC, MPCBatchedState, MPCInput, MPCSingleState
 
 
 class Task(ABC):
-    """A task is a wrapper around an gym environment and an MPC planner."""
+    """A task describes a concrete problem to be solved by a learning problem.
+
+    This class serves as a base class for tasks that involve a combination of
+    a gymnasium environment and a model predictive control (MPC) planner. It
+    provides an interface for preparing neural network inputs and MPC inputs
+    based on environment observations and states.
+
+    Attributes:
+        mpc (MPC): The Model Predictive Control planner to be used for this task.
+        env (gym.Env): The gymnasium environment for the task.
+    """
+
     def __init__(self, mpc: MPC, env: gym.Env):
-        """A task is a wrapper around an gym environment and an MPC planner.
+        """Initializes the Task with an MPC planner and a gymnasium environment.
 
         Args:
-            mpc: The learnable MPC planner.
-            env: The gym environment.
+            mpc (MPC): The Model Predictive Control planner to be used for this task.
+            env (gym.Env): The gymnasium environment for the task.
         """
         super().__init__()
         self.mpc = mpc
         self.env = env
 
     @abstractmethod
-    def prepare_nn_input(self, obs: Any) -> torch.Tensor:
-        """Prepare the neural network input from the observation.
+    def prepare_nn_input(self, obs: Any) -> np.ndarray:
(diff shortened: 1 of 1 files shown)


leap_c/task.py
```diff
@@ -0,0 +1,44 @@
...
+        """Prepare the neural network input from the observation.
+
+        Args:
+            obs: The observation from the environment.
+        """
+        ...
+
+    @abstractmethod
+    def prepare_mpc_input(
+        self, mpc_state: MPCBatchedState, obs: Any, param_nn: Optional[torch.Tensor]
```
Confectio, line 34: I think we should allow the types "MPCSingleState | MPCBatchedState | None = None" as in __call__ for mpc_state here.
Or maybe I misunderstand the intention of this mpc_state?

leap_c/task.py
```diff
@@ -0,0 +1,44 @@
...
+from typing import Any, Optional
+
+import gymnasium as gym
+import torch
+
+from leap_c.mpc import MPC, MPCBatchedState, MPCInput
+
+
+class Task(ABC):
+    """A task is a wrapper around an gym environment and an MPC planner."""
```
Confectio, line 11: Maybe add that it is ment to represent a concrete learning task?

leap_c/task.py
```diff
@@ -0,0 +1,44 @@
...
+    def __init__(self, mpc: MPC, env: gym.Env):
+        """A task is a wrapper around an gym environment and an MPC planner.
```
Confectio, line 13: Should this really be the same as the class description? I forgot what exactly our policy for docstring in class vs docstring in __init__ was :)

leap_c/task.py
```diff
@@ -0,0 +1,44 @@
...
+        ...
+
+    @abstractmethod
+    def prepare_mpc_input(
+        self, mpc_state: MPCBatchedState, obs: Any, param_nn: Optional[torch.Tensor]
+    ) -> MPCInput:
+        """Prepare the MPC input from the observation.
+
+        Args:
+            mpc_state: The current state of the MPC.
```
Confectio, line 39: I think this shouldn't be the current state of the mpc but rather the state corresponding to the observation or something.

Confectio (changes requested):
Looks good in general, but I have some questions/remarks that should be resolved :) .

leap_c/task.py
JasperHoffmann: Added both options, in the future we should just default to batched input. Makes our life significantly easier ;)

leap_c/task.py
```diff
@@ -0,0 +1,67 @@
...
+        """
+        ...
+
+    @abstractmethod
+    def prepare_mpc_input(
+        self,
+        mpc_state: MPCSingleState | MPCBatchedState,
+        obs: Any,
+        param_nn: Optional[np.ndarray],
+    ) -> MPCInput:
```
Confectio, line 55: Now that I think about it, shouldn't we also output an mpc_state here? If not, what was our intention with the input mpc_state?

Confectio (changes requested):
Looks very nice now :D . I still have a question, though.

leap_c/task.py
JasperHoffmann: The mpc_state is part of MPCInput?

Confectio: It is not, as far I can see, only x, u, and MPCParameters are part of MPCInput

Confectio: But I am not against making it a part of MPCInput, if that is what you are trying to say.

JasperHoffmann: Ah yes, we can remove the mpc_state! Not sure what the thought process was here!

Confectio: Maybe we had some specific thought, since mpc_state is an input in our schematic from the whiteboard, but right now I also do not know what we had in mind 🤷 .

Confectio (approved):

PR #11, opened 2025-02-03
Collect basic stats in each MPC solve 

Confectio:
This adds the collection of some potentially interesting MPC solver statistics.

Note that
1. I think it is a little bit annoying to save these stats in a member of the MPC, but it seems to me that this is better than blowing up the output of the MPCLayer with Tensors for every statistic that is interesting
(note that autograd does not allow other outputs than Tensors).
2. We have to collect them in the call (not afterwards only at request), because of the potential re-solve, which would override stats in the solver.

### DIFF
--- a/leap_c/mpc.py
+++ b/leap_c/mpc.py
@@ -340,12 +340,12 @@ def _solve_shared(
     # TODO: Cover case where we do not want to do a forward evaluation
     solver.solve()
 
-    solve_infos = dict()
+    solve_stats = dict()
 
     if isinstance(solver, AcadosOcpSolver):  #
-        solve_infos["sqp_iter"] = solver.get_stats("sqp_iter")
-        solve_infos["qp_iter"] = solver.get_stats("qp_iter").sum()  # type:ignore
-        solve_infos["time_tot"] = solver.get_stats("time_tot")
+        solve_stats["sqp_iter"] = solver.get_stats("sqp_iter")
+        solve_stats["qp_iter"] = solver.get_stats("qp_iter").sum()  # type:ignore
+        solve_stats["time_tot"] = solver.get_stats("time_tot")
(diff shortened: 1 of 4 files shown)


leap_c/mpc.py
```diff
@@ -351,10 +356,31 @@ def _solve_shared(
...
-            any_failed = False
+                solve_stats["first_solve_status"] = solver.status
+            else:
+                solve_stats["first_solve_status"] = 0
+        solve_stats["sqp_iter"] += solver.get_stats("sqp_iter")
+        solve_stats["qp_iter"] += solver.get_stats("qp_iter").sum()  # type:ignore
+        solve_stats["time_tot"] += solver.get_stats("time_tot")
+
+    elif isinstance(solver, AcadosOcpBatchSolver):
+        status_batch = []
```
dirkpr, line 367: With a fixed batch size n_batch, we could pre-allocate the memory for the stats instead of appending to empty lists. Or use list comprehension.

dirkpr (commented):
Looks good. Only one minor comment.

leap_c/mpc.py
Confectio: Usually I avoid list comprehension when I would have to iterate over the same thing a few times (e.g., If I'd write a list comprehension for every stat, I would do 4 loops over the same thing instead of one loop where I collect everything I want at once). Not sure if one or the other thing is actually better performance-wise, but I think the difference is negligible? Or would you argue list comprehension is better for readability (I am also not sure about that, I feel like it is not, but maybe its just what I am more used to)?

I agree with the preallocation, but I hope that we will have the possibility of variable batch sizes in the not-too-far future and I hence think that changing this (probably comparably small) inefficiency is not worth it :) .

PR #12, opened 2025-02-05
Cartpole upgrade (GN upgrade)

Confectio:
Upgrades the Cartpole environment and the MPC class such that Linear Least Squares cost can be used in the first solver (enabling, e.g., GN Hessian approximation)
In particular
- More general costs (LLS or general quadratic) and more flexible parameterization for it.
- Extended the Cartpole Environment such that LLS costs can be used
- Extended the MPCParameter interface to also incorporate LLS fields 
(as far as I see in the acados docs, this can also be used already for NONLINEAR_LS and CONVEX_OVER_NONLINEAR cost_types)
and implemented the necessary functionality
- Docstrings for the pendulum MPC :)

Other:
- A little refactoring
- New Tests and fixtures.
- Bugfixes (in particular fixes accidentally using the normal solver instead of the sensitivity solver for sensitivities dudx0)

### DIFF
--- a/leap_c/examples/pendulum_on_cart.py
+++ b/leap_c/examples/pendulum_on_cart.py
@@ -63,21 +63,15 @@ class PendulumOnCartMPC(MPC):
                 "uref": np.array([0]),  # reference position, only used for LS cost
             }
 
-        ocp = export_parametric_ocp(
+        ocp, ocp_sens = export_parametric_ocp(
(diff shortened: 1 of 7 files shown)


JasperHoffmann (approved):
I think the changes are good. Maybe we should discuss the default param feature again. Do we still need this or is this covered by the default_init_fn?

PR #13, opened 2025-02-10
Pointmass example

dirkpr:
Implement 2-dim pointmass example in (mpc,env,task) structure. The goal is to drive the pointmass to the origin by applying force inputs. A random disturbance acts along the direction of the action and the reference is on the constraint boundary. The idea is that the agent learns to align the action close to the origin such that the noise does not push it over the constraint.

Also implement LinearMPC class, but using MPC class in the pointmass mpc for now. Extend to LinearMPC in next PR. The idea is to compare timings.


JasperHoffmann (approved):
Looks good!

PR #14, opened 2025-02-12
Add new SAC trainer

Confectio:
Large refactoring of trainer class to provide close to single-file implementations for each RL algorithm.

Broadly, this PR implements
- SAC trainer that does not use the MPC
- SAC_FOU to use MPC parameterized by SAC
- Implroved Logging
- Improved ReplayBuffer
- Refactor stochastic policy functionality
- Results after debugging SAC


dirkpr (approved):
Looks good! Collective review by Confectio, Jasper and Dirk offline.

PR #16, opened 2025-02-19
Added propper seeding

JasperHoffmann:
Added proper seeding:
- Added seeding to envs in task for observation spaces and action spaces.
- Seeded numpy torch gymnasium and the python random module
- Also updated the main run script to be more applicable to be used from other scripts

### DIFF
--- a/leap_c/utils.py
+++ b/leap_c/utils.py
@@ -194,6 +194,5 @@ def set_seed(seed: int):
     random.seed(seed)
     np.random.seed(seed)
     torch.manual_seed(seed)
-    torch.cuda.manual_seed_all(seed)
     torch.backends.cudnn.deterministic = True
     torch.backends.cudnn.benchmark = False
--- a/leap_c/utils.py
+++ b/leap_c/utils.py
@@ -2,6 +2,7 @@ import atexit
 import os
 import shutil
 from pathlib import Path
+import random
 from tempfile import mkdtemp
 from typing import Any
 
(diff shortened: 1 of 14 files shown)


run.py
```diff
@@ -13,51 +15,64 @@
...
+        task_name: Name of the task to use.
+        cfg: The configuration to use.
+        output_path: Path to save output to.
+        device: Device to run on.
+    """
+    if output_path.exists():
+        raise ValueError(f"Output path {output_path} already exists")
 
     task = create_task(task_name)
     cfg = create_default_cfg(trainer_name)
```
Confectio, line 69: I think this line contains a bug. It always overrides the given cfg argument with the default cfg.

leap_c/utils.py
```diff
@@ -186,3 +187,14 @@ def set_standard_sensitivity_options(ocp_sensitivity: AcadosOcp):
...
     ocp_sensitivity.model.name += "_sensitivity"  # type:ignore
+
+
+def set_seed(seed: int):
+    """Set the seed for all random number generators."""
+    random.seed(seed)
+    np.random.seed(seed)
+    torch.manual_seed(seed)
+    torch.cuda.manual_seed(seed)
+    torch.cuda.manual_seed_all(seed)
```
Confectio, line 198: I think torch.manual_seed already covers torch.cuda.manual_seed_all (and hence, also torch.cuda.manual_seed) ? See https://github.com/pytorch/pytorch/blob/v2.6.0/torch/random.py#L32

Confectio (changes requested):
I think I spotted a bug, but aside from this it looks nice :)

leap_c/utils.py
JasperHoffmann: Removed 197 and 198.

Confectio (approved):
👍

PR #17, opened 2025-02-25
Logging and some bugfixes

Confectio:
Main features of this PR:
- Bugfixes (medium bug in mpc, small one in log prob)
- Extract some more information and log it in SAC_fop

(pls don't delete the branch after this pr is merged, I will continue here)


JasperHoffmann (approved):
This should work!

Maybe when we refactor the MPCLayer, we can make it such that the stats are provided completely by the MPCLayer. Currently, the stat generation for this is distributed at a lot of locations....

JasperHoffmann (approved):
Looks good!

PR #18, opened 2025-02-27
General improvements

Confectio:
In particular, 
fixed some logging introduced in the last PR, 
improved wandb logging, 
set warmstarting options in ocp solver,
and updated acados(!)


leap_c/trainer.py
```diff
@@ -169,7 +173,14 @@ def __init__(
 
         # init wandb
         if cfg.log.wandb_logger:
-            wandb.init(project="leap", dir=self.output_path / "wandb")
+            wandbdir = self.output_path / "wandb"
+            wandbdir.mkdir(exist_ok=True)
+            wandb.init(
+                project="leap",
```
JasperHoffmann, line 179: Should this also be added to config?

Confectio: Yeah, I think it would be better.

leap_c/rl/sac_fop.py
```diff
@@ -57,26 +57,39 @@ def put_status_into_stats(status: np.ndarray | torch.Tensor, stats: dict, prefix
...
+    first_solve_status = mpc_stats.pop("first_solve_status", None)
+    if first_solve_status is not None:
+        put_status_into_stats(
+            status=first_solve_status, stats=mpc_stats, prefix="first_solve_"
+        )
     put_status_into_stats(status=actual_status, stats=mpc_stats, prefix="actual_")
-    sum_up_dict(mpc_stats, mpc_stats_summed_up)
+    for k in mpc_stats.keys():
+        mpc_stats_summed_up[k] = mpc_stats_summed_up.get(k, 0) + mpc_stats[k]
+        if k in MPC_STATS_NON_STATUS_KEYS:
```
JasperHoffmann, line 74: I think is already a  lot of acados specific code. We should move this into the class MPC :) Will try to make a update on this next week!

JasperHoffmann (approved):
Looks good. Though I am not completely happy with the logging of the MPC stats!

PR #20, opened 2025-03-04
Zero order

JasperHoffmann:
Added zero order mpc-sac implementation:
- Updated sac, sac_fop, sac_zop to be more consistent
- Updated general logging
- Removed action logging for now!


leap_c/mpc.py
```diff
@@ -425,7 +438,12 @@ def _solve_shared(
             solve_stats["sqp_iter"] += solver.get_stats("sqp_iter")
             solve_stats["qp_iter"] += solver.get_stats("qp_iter").sum()  # type:ignore
             solve_stats["time_tot"] += solver.get_stats("time_tot")
-        solve_stats["first_solve_status"] = [solver.status]
```
Confectio, line 428: I would also log the first solve status, not remove it (note that this is bugged here, this should be the status before the reattempt)

leap_c/rl/sac_fop.py
```diff
@@ -224,42 +182,37 @@ def init_policy_state(self) -> Any:
...
                 policy_state = self.init_policy_state()
-                mpc_stats_aggregated_rollout = {}
                 is_terminated = is_truncated = False
                 episode_return = episode_length = 0
-            action, policy_state_prime, param, status, mpc_stats = self.act(
-                obs, state=policy_state
-            )
+                episode_act_stats = defaultdict(list)
+
+            action, policy_state_prime, act_stats = self.act(obs, state=policy_state)
```
Confectio, line 205: I would prefer this looked like in sac_zop, it is more consistent that way and information is more accessible, in case one needs to do something with it.

leap_c/mpc.py
Confectio: But I strongly agree that the status logging should happen here in the MPC, I don't know why I didn't think of it myself so far 👍 .

leap_c/nn/modules.py
```diff
@@ -42,6 +42,52 @@ def forward(
...
+#     loss_stats: dict,
+#     actual_status: np.ndarray | torch.Tensor,
+# ):
+#     first_solve_status = mpc_stats.pop("first_solve_status", None)
+#     if first_solve_status is not None:
+#         put_status_into_stats(
+#             status=first_solve_status, stats=mpc_stats, prefix="first_solve_"
+#         )
+#     put_status_into_stats(status=actual_status, stats=loss_stats, prefix="actual_")
+#     loss_stats.update(mpc_stats)
```
Confectio, line 88: I think we should not leave code laying around like this, rather remove it.

leap_c/rl/sac_fop.py
```diff
@@ -153,24 +106,27 @@ def forward(self, obs, mpc_state: MPCBatchedState, deterministic=False):
...
-            )
+        # if action.shape[0] == 1:
+        #     stats = {
+        #         param_labels[k]: element for k, element in enumerate(param.squeeze())
+        #     }
+        #     self.trainer["trainer"].report_stats(
+        #         "action",
+        #         stats,
+        #         self.trainer["trainer"].state.step,
+        #     )
```
Confectio, line 119: I think we should not leave code laying around like this, rather remove it (but please bring back Parameter logging asap :) ).

leap_c/rl/sac_zop.py
```diff
@@ -135,25 +131,31 @@ def forward(self, obs, mpc_state: MPCBatchedState, deterministic=False):
...
-            )
+        # if action.shape[0] == 1:
+        #     stats = {
+        #         param_labels[k]: element for k, element in enumerate(param.squeeze())
+        #     }
+        #     self.trainer["trainer"].report_stats(
+        #         "action",
+        #         stats,
+        #         self.trainer["trainer"].state.step,
+        #     )
```
Confectio, line 144: I think we should not leave code laying around like this, rather remove it (but please bring back parameter logging asap :) ).

leap_c/run.py
```diff
@@ -75,6 +75,9 @@ def main(
         cfg=cfg,
     )
 
+    cfg.val.interval = 2000
+    cfg.val.num_render_rollouts = 1
```
Confectio, line 79: I think this shouldn't stay here?

leap_c/task.py
```diff
@@ -23,29 +24,39 @@ class Task:
...
-                vironment for the task.
+            collate_fn_map (dict[type, Callable]): A dictionary mapping types to collate
+                functions. If None, the default collate function map is used.
         """
         super().__init__()
         self.mpc = mpc
-        self.env_factory = env_factory
-        self.collate_fn_map = create_collate_fn_map()
-        self._seed = None
+        self.collate_fn_map = (
```
Confectio, line 46: I feel like this collation map is something that the algorithm itself should take care of (since the algorithm is also the one that knows what is being put into the buffer), but we can also try this🤷‍♂️ .

leap_c/env_wrapper.py
```diff
@@ -0,0 +1,18 @@
...
+
+from typing import Any, SupportsFloat
+import gymnasium as gym
+
+
+class ActionStatsWrapper(gym.Wrapper):
+    """A wrapper that logs the actions taken by the agent."""
+
+    def step(self, action) -> tuple[Any, SupportsFloat, bool, bool, dict[str, Any]]:
+        *transition, info = super().step(action)
```
Confectio, line 11: Very nice syntax, I oftentimes needed something like this, but didn't bother to look it up. Thanks, definitely will try :D

leap_c/trainer.py
```diff
@@ -53,6 +54,8 @@ class LogConfig:
 
     val_window: int = 1
 
+    log_actions: bool = False
```
Confectio, line 57: Did you try it out, does it work? 
As far as I can see, the actionwrapper only modifies the info from the environment, and we always ignore the info passed from the environment, in training and in validation :') .

Confectio (changes requested):
Nice PR, but I have some remarks :) .

leap_c/trainer.py
JasperHoffmann: I just added it for future use ;) First want to think about how to log such stuff in a general way.

leap_c/task.py
JasperHoffmann: I think this is the easiest as the task might have different objects that need to be considered when doing a collate :)

leap_c/run.py
JasperHoffmann: Yes can remove it!

leap_c/task.py
Confectio: Can you give an example?

JasperHoffmann: E.g. a spline function!

JasperHoffmann: However, we might be able to simplify the collate fn such that objects that are unknown will be not stacked and just put in a list. I could see this a a potential simplification!

leap_c/nn/modules.py
JasperHoffmann: Removed!

leap_c/mpc.py
```diff
@@ -461,7 +479,14 @@ def _solve_shared(
                 sqp_iter_batch[i] += ocp_solver.get_stats("sqp_iter")
                 qp_iter_batch[i] += ocp_solver.get_stats("qp_iter").sum()  # type:ignore
                 time_tot_batch[i] += ocp_solver.get_stats("time_tot")
-            solve_stats["first_solve_status"] = status_batch
+
+        # report each status individually
+        status_batch = np.array(status_batch)
```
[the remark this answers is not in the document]
JasperHoffmann: Yes, I added a ticket to improve status logging in the future, such that we also track how many retries we need and report the status after the second round.

leap_c/rl/sac_zop.py
JasperHoffmann: True, I will leave it for now! And we will bring it back in another commit!

dirkpr (approved):
Looks good functionality-wise. Can do some refactoring in a later PR.

leap_c/rl/sac_fop.py
```diff
@@ -23,105 +25,58 @@
...
+            Callable[[torch.Tensor, torch.Tensor, MPCBatchedState], MPCBatchedState]
+            | None
+        ) = None,
     ):
         super().__init__()
 
         param_space = task.param_space
 
-        self.extractor = task.create_extractor()
+        self.extractor = task.create_extractor(env)
```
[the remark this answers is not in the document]
JasperHoffmann: Yes, I was also thinking about passing a extractor factory instead, but thought this is somehow easier ^^

PR #22, opened 2025-03-06
Improved docstrings in task descriptions.

JasperHoffmann:
Improved docstrings of Tasks.

Atomic commits 🗡️

### DIFF
--- a/leap_c/task.py
+++ b/leap_c/task.py
@@ -26,7 +26,9 @@ class Task(ABC):
         mpc (MPC): The Model Predictive Control planner to be used for this task.
         collate_fn_map (dict[type, Callable]): A dictionary mapping types to collate
             functions. This is used to collate data into a tensor. If None, the default
-            collate function map is used, which is often sufficient for most tasks.
+            collate function map is used, which is sufficient for most tasks and contains
+            some extensions to the default PyTorch collate function to handle acados
+            objects.
     """
 
     def __init__(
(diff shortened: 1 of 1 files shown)


leap_c/task.py
```diff
@@ -25,7 +25,8 @@ class Task(ABC):
     Attributes:
         mpc (MPC): The Model Predictive Control planner to be used for this task.
         collate_fn_map (dict[type, Callable]): A dictionary mapping types to collate
-            functions.
+            functions. This is used to collate data into a tensor. If None, the default
+            collate function map is used, which is often sufficient for most tasks.
```
Confectio, line 29: I would write "which is sufficient for most tasks" (leave often).
We should also make explicitly clear here that we are using OUR default collate function which entails more than pytorch's default collate. Also maybe it would be nice to refer to https://pytorch.org/docs/stable/data.html#torch.utils.data.default_collate for people who do not know collate.

leap_c/task.py
```diff
@@ -39,7 +40,8 @@ def __init__(
             mpc (MPCSolutionModule): The Model Predictive Control planner to be used
                 for this task.
             collate_fn_map (dict[type, Callable]): A dictionary mapping types to collate
-                functions. If None, the default collate function map is used.
+                functions. If None, the default collate function map is used, which is
+                often sufficient for most tasks.
```
Confectio, line 44: I would write "which is sufficient for most tasks" (leave often).
We should also make explicitly clear here that we are using OUR default collate function which entails more than pytorch's default collate. Also maybe it would be nice to refer to https://pytorch.org/docs/stable/data.html#torch.utils.data.default_collate for people who do not know collate.

Confectio (changes requested):
I might have a few more improvements :)

leap_c/task.py
JasperHoffmann: Good Idea!

Confectio (approved):

PR #21, opened 2025-03-05
General improvements

Confectio:
Main points:
- Bugfix for the default function being used despite being None 
- Bugfix for log prob prime
- Bugfix using constant zeroes iterate despite iterate being given 
- A lot of renaming (e.g., MPC -> Mpc)
- Improved logging of mpc stats (status)
- Remastered and reactivated the frozen mpc tests

PR #24, opened 2025-03-06
Bugfix target calculation

Confectio:
The target calculation has been messed up in #21 (wrong shapes), this fixes it again (also fixes another bug that only appears when trying to use prepare_mpc_state in sac_fop's MpcSacActor).


JasperHoffmann:
Perfect!

JasperHoffmann (approved):
Looks good!

PR #34, opened 2025-03-10
Logging

JasperHoffmann:
Improved logging:
- Improved environment logging by making it more consistent and adding entropy as well as the unsquashed std deviation.
- Added back the SquashedGaussian to make the code easier to maintain.
- Added a default init alpha to speed up convergence for environments that do not require a lot of initial exploration
- Changed the entropy calculation of FOP and ZOP to relate to the action space dims instead of the param dims: We normalize the entropy bonus by multiplying it with |A| / |P|. This leads to drastically faster convergence of all algorithms!
- Improved device management, we load the network to the device in trainer.run


dirkpr (approved):
Looks good!

PR #35, opened 2025-03-11
Adding a drone example

RudolfReiter:
The drone environment involves stabilizing a quadrotor from a random position on a sphere at the origin. The MPC horizon is kept very short. Two variants of the task are contained: 1. varying the mass of the drone in a closed-loop and 2. varying the terminal reference in the closed-loop.
An evaluation script is added to test the MPC implementation.


scripts/rr/run_qr.py
JasperHoffmann, line 1: Can we name the folder quadrotor?

JasperHoffmann (approved):
Looks good, I think we should name the script folder differently :)

scripts/rr/run_qr.py
JasperHoffmann: Just renamed it :)

PR #36, opened 2025-03-11
Orthogonal init for MLPs

Confectio:
- Fix possible bug (shape in policy loss).
- Add orthogonal init to MLPs


leap_c/nn/mlp.py
JasperHoffmann, line 1: Can we make the weight_init an argument here, which can be None and then orthogonal weight initialization is used.?

leap_c/rl/sac.py
```diff
@@ -226,7 +226,7 @@ def train_loop(self) -> Iterator[int]:
 
                 # update actor
                 q_pi = torch.cat(self.q(o, a_pi), dim=1)
-                min_q_pi = torch.min(q_pi, dim=1).values
+                min_q_pi = torch.min(q_pi, dim=1, keepdim=True).values
```
JasperHoffmann, line 229: looks good!

leap_c/rl/sac_fop.py
```diff
@@ -133,6 +135,7 @@ def forward(self, obs, mpc_state: MpcBatchedState, deterministic=False):
         # TODO: We have to catch and probably replace the state_solution somewhere,
         #       if its not a converged solution
         mpc_output, state_solution, mpc_stats = self.mpc(mpc_input, mpc_state)
+        self.actual_used_mpc_state = mpc_state
```
JasperHoffmann, line 138: Why should we save it like this? Are you using this in your repo?

JasperHoffmann (changes requested):
Looks good, I am not completely sure why we should already save the state in the actor module though.

leap_c/nn/mlp.py
Confectio: Yes, will do.

leap_c/rl/sac_fop.py
Confectio: Yes. 
The problem: With prepare_mpc_state we introduced a possibility to use an other mpc state than we actually put into the forward method. This means, from the outside (in the training loop), we do not know how the state that we really used actually looked like. 
The (possible) solution: Save the mpc state that has actually been used in a member, so it can be retrieved in case people want to know it (E.g., I log the distance of this state to the actual solution state).

JasperHoffmann: I would find it cleaner if you put into the output tuple ;)

JasperHoffmann (approved):

PR #37, opened 2025-03-11
Add flag to turn off entropy reward bonus

JasperHoffmann:
Add flag in the sac implementations to turn off the reward bonus.


Confectio (approved):
Looks good :)

PR #38, opened 2025-03-11
Fixed q report bug

JasperHoffmann:
Fixed a bug that we reported an average parameter of the q-network instead of the real q values.


Confectio (approved):
Nice catch ^^

PR #40, opened 2025-03-12
Improved mpc stats logging.

JasperHoffmann:
This PR makes the mpc stats logging more consistent. For a future state we could try to reduce the number of logged metrics or have a better description of logged metrics.

### DIFF
--- a/leap_c/mpc.py
+++ b/leap_c/mpc.py
@@ -1,5 +1,6 @@
 from abc import ABC
 from copy import deepcopy
+from collections import defaultdict
 from dataclasses import fields
 from enum import IntEnum
 from functools import cached_property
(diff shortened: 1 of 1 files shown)


leap_c/mpc.py
```diff
@@ -435,44 +446,44 @@ def _solve_shared(
                         set_params=False,
                         throw_error_if_u0_is_outside_ocp_bounds=throw_error_if_u0_is_outside_ocp_bounds,
                     )
+            # TODO (Jasper): We also resolve when it already converged...
```
Confectio, line 449: Yes, but the solution iterate is already in the respective solver, so the solver notices it is already at the solution and ceases calculations almost immediately. Since this happens in parallel to actually solving a not-converged sample I would expect this to be negligible.

Confectio (approved):
Nice, I like it 👍 . Concerning the future logging of status, at least from my experience so far, I only encountered status 0, 1, 2, 4.

leap_c/mpc.py
JasperHoffmann: Yes, I also checked the performance of the backup_fn and it is not really a problem performance wise. The main bottleneck of the batch solver is the number of threads in your system.

PR #43, opened 2025-03-13
Critic input normalization

JasperHoffmann:
We added some improvements to all sac implementations and sac_fop:
- Normalizing actions for the critic input for sac, sac_zop and sac_fop
- Made the masking of samples stricter. The mpc needs to converge for the current action and next state action.
- Added three example scripts for the quadrotor task.
- Removed unnecessary code in `leap_c/nn/utils.py`


leap_c/rl/sac_fop.py
```diff
@@ -242,6 +255,25 @@ def train_loop(self) -> Iterator[int]:
 
                 # sample action
                 pi_o = self.pi(o, ps_sol)
+                with torch.no_grad():
+                    pi_o_prime = self.pi(o_prime, ps_sol)
+
+                pi_o_stats = pi_o.stats
+
+                # compute mask
+                mask_status = (pi_o.status == 0) & (pi_o_prime.status == 0)
```
dirkpr, line 264: Should we define a mask somewhere else and configure the acceptable status here? Max iterations is not status 0 for instance.

leap_c/nn/utils.py
```diff
@@ -70,33 +18,36 @@ def string_to_activation(activation: str) -> nn.Module:
...
-        fcs.extend([nn.Linear(prev_d, d), string_to_activation(activation)])
-    if use_activation_on_output:
-        mlp = nn.Sequential(*fcs)
-    else:
-        mlp = nn.Sequential(*fcs[:-1])
 
-    # TODO(Leonard): Do we need orthogonal initialization?
-    # mlp.apply(weight_init)
-    return mlp
+def normalize(
```
dirkpr, line 34: min_max_scaling would be more specific, but I guess this is accepted ML lingo

dirkpr (approved):
Looks good. Only two minor comments.

leap_c/nn/utils.py
JasperHoffmann: I think this a good idea :)

leap_c/rl/sac_fop.py
JasperHoffmann: Yes, sounds great, we can also even allow in the config to determine, which stati should be filtered out :)

Confectio: But maybe we need to make it even more general than just specifying which status are ok? E.g., max iterations can mean the sum of residuals still being >> 1 (not ok?), but it could also mean the sum of residuals being at 1e-4 (ok?).

dirkpr: I would just mask for status at this point and have user specify acceptable solver tolerance in the configure_ocp_solver function

JasperHoffmann: Yes, I think this easies to implement as well :)

JasperHoffmann: I decided against implementing it. I think this is an Acados Specific detail that should not be handled in the sac but in the MpcLayer :)

PR #44, opened 2025-03-13
Small bugfix in chain env

Confectio:
Return type of reward and terminated was ca.DM, which caused the collate to fail.


dirkpr (approved):
Good catch of the bug! I also see we need a ruff.toml

Confectio:
Ah, indeed. I thought everyone was just using the same "standard" ruff configuration that is set when ruff is installed.

PR #45, opened 2025-03-13
Critic input normalization

JasperHoffmann:
Renamed normalization to min_max_scaling and added action normalization for the Zop critic.


Confectio (approved):
😄

PR #48, opened 2025-03-13
Fix logging masked samples

Confectio:
Logging masked samples contained a bug, because the status has been removed from pi_o before via pi_o.select(mask_status).


JasperHoffmann (approved):
Good catch!

PR #50, opened 2025-03-18
Bugfix seeding in environments

Confectio:
The pointmass environment and the chain environment had random elements that were not being seeded properly. 
This was not being noticed, because the tests checking this were not being used on those environments. This PR fixes both.

Also comes with a little refactoring and a chain Task for learning the mass parameters (only 2 parameters to learn).


dirkpr (approved):
Looks good!

PR #52, opened 2025-03-31
Make Parallelization more accessible

Confectio:
The number of threads to use for parallelizing the mpc batch calculations previously had to be set in the ocp formulation. 
It can now be set dynamically at any point in time, which is exposed in the Mpc class by the property `num_threads_in_batch_methods` and the equally named constructor argument.

Further, it was actually not set at all by default, despite large computational speedups being enabled by it.
Hence, there is now a new field `num_threads_in_mpc_batch_methods` in the SacFopAlgorithmConfig (SacFop is currently the only architecture using mpc batch calculations).
It is set to  `4` threads by default.

Enabled by: [acados/acados/#1483](https://github.com/acados/acados/pull/1483)

### DIFF
--- a/leap_c/rl/sac_fop.py
+++ b/leap_c/rl/sac_fop.py
@@ -30,6 +30,7 @@ class SacFopAlgorithmConfig:
         critic_mlp: The configuration for the critic networks.
         actor_mlp: The configuration for the policy network.
         batch_size: The batch size for training.
+        num_threads_in_mpc_batch_methods: The number of threads to use for mpc batch methods.
         buffer_size: The size of the replay buffer.
         gamma: The discount factor.
         tau: The soft update factor.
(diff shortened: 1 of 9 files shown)


leap_c/rl/sac_fop.py
JasperHoffmann, line 1: `num_threads_in_mpc_batch_methods`

I find the name a little bit too long! Maybe `num_threads_mpc_solver`. We ideally set the number of threads in the MPC class, as this is an implementation detail of acados. Maybe we could make it a property there, such that people can overwrite it in a train script :)

JasperHoffmann: As this can be only set once currently? We could make it such that it needs to be set before the solver is created the first time? It is probably anyways already a advanced setting?

Confectio: The name: Yeah, it is a little long, we could even change it to num_threads_mpc if you want. On the other hand, I think it doesn't hurt when people know exactly what it paralellizes, just by adding a few words to this option (its not like people have to type it all the time, since they probably set it only rarely)?
Also the name is very similar to the acados name.

Ive already added properties in the MPC class, so this can be set at any time people want 👍

Confectio:
I already added it to the sac fop config, because the sac fop architecture is actually the only architecture so far using this paralellization. Where would you set it otherwise?

leap_c/rl/sac_fop.py
JasperHoffmann: Yes, let's go with num_threads_mpc!

JasperHoffmann (approved):
Apart from the naming it looks good!

PR #53, opened 2025-04-01
Automatic build of docs

josipkh:
This PR adds a GitHub workflow to automatically build the documentation stored in `docs/source`.
The output is stored on a separate branch named `gh-pages`, which will be used to generate the page.
The workflow is based on [this exercise](https://coderefinery.github.io/documentation/gh_workflow/).

Building the docs could also be triggered only in specific cases, for now it is set to basically every change to the repository.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -1,0 +1,27 @@
+name: documentation
+
+on: [push, pull_request, workflow_dispatch]
+
+permissions:
+  contents: write
+
+jobs:
+  docs:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


josipkh:
Possible [solution](https://github.com/ad-m/github-push-action/issues/96#issuecomment-1647904286) to the failing workflow

josipkh:
Workflows from PRs coming from public forks are not allowed to write content ([link](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token)).

PR #51, opened 2025-03-24
Gauss-Newton with external cost

FreyJo:
# Support for Gauss-Newton Approximations with Linear and Nonlinear LS Cost Types

## 🔑 Key Improvements
- Added support for parametric `LINEAR_LS` and `NONLINEAR_LS` cost with `p_global values` (translated to EXTERNAL)
- Unified cost update interface for both forward and sensitivity solvers
- Removed special treatment requirements in `MpcParameter`
- Enabled Gauss-Newton Hessian approximations in forward solver with exact Hessians in sensitivity solver. Using the same interface to update cost parameters
- Avoid expensive calls to `cost_set`

## 🛠️ Technical Changes
- Refactored cost functions and sensitivity options in core `MPC` class
- Modified pendulum on cart MPC example to utilize `NONLINEAR_LS` cost type
- Renamed and removed obsolete parameters
- Updated test fixtures and test functions for compatibility

## 💡 Value
This enhancement provides a more elegant and consistent approach to cost handling. By translating `NONLINEAR_LS` cost to `EXTERNAL` cost and implementing a unified interface for updates via `p_global_values`, we maintain the computational efficiency of Gauss-Newton approximations while ensuring sensitivity calculations remain exact through exact Hessians.

---
*Enabled by: https://github.com/acados/acados/pull/1476*

PR #54, opened 2025-04-01
Fix cartpole tasks

Confectio:
The cartpole swingup task was actually a cartpole balancing task. 
Fixed it and added a balancing task.  
Further, removes the now unnecessary setting of the references, as described in #51 .


JasperHoffmann (approved):
Looks good! Have you tried training on it again?

Confectio:
I tried training at the "old" environment (which had the name swingup but was balance), which worked. I reran it shortly after doing the changes and it seemed to train, so I was satisfied for now. But I will do a whole run soon :)

PR #55, opened 2025-04-02
Add documentation workflow

dirkpr:
Workflow to build `docs` and host it on [github](https://leap-c.github.io/leap-c/)


.github/workflows/documentation.yml
```diff
@@ -0,0 +1,78 @@
...
+     shell: bash
+     run: |
+       .github/linux/install_tera.sh
+   - name: Export Paths
+     working-directory: ${{github.workspace}}
+     shell: bash
+     run: |
+       echo "ACADOS_SOURCE_DIR=${{github.workspace}}/leap_c/external/acados" >> $GITHUB_ENV
+       echo "ACADOS_INSTALL_DIR=${{github.workspace}}/leap_c/external/acados" >> $GITHUB_ENV
+       echo "LD_LIBRARY_PATH=${{github.workspace}}/leap_c/external/acados/lib" >> $GITHUB_ENV
```
FreyJo, line 39: I would guess all the 3 acados specific steps above are not needed to build docs, or are they?

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,78 @@
...
+     uses: actions/cache@v4
+     with:
+       path: ~/.cache/pip
+       key: ${{ runner.os }}-pip-${{ hashFiles('**/pyproject.toml') }}
+       restore-keys: |
+         ${{ runner.os }}-pip-
+   - name: Install new Casadi
+     working-directory: ${{github.workspace}}/leap_c
+     run: |
+       ./install_new_casadi_py311_x86_64.sh
```
FreyJo, line 55: Not needed anymore, casadi 3.7 is out now and will be installed automatically in the next step.

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,78 @@
+name: documentation
+
+on: [push, pull_request, workflow_dispatch]
```
FreyJo, line 3: Can we only deploy on commits in `main`?
It seems that now open PRs are just uploading/overwriting the docs page.

FreyJo (commented):
Great stuff!
I think we could avoid a few steps in the build process.

.github/workflows/documentation.yml
dirkpr: Removed in https://github.com/leap-c/leap-c/pull/55/commits/96f1c4cb3329478036333d16dd55f22ce45ad7bc

.github/workflows/documentation.yml
dirkpr: I agree on first glance. The problem is that we use `automodule` in documenting files like here https://github.com/leap-c/leap-c/tree/9b6c9aece0c148bb3150824e2c3e3c6382d5d433/docs/source/api , where we need to import `acados`. We could either try living without `automodule` or append the building of the docs to a workflow where `acados` already exists.

.github/workflows/documentation.yml
josipkh: agreed, changed

dirkpr: Changed it to `on: workflow_dispatch`, in case one wants to work on the documentation in a separate branch and test

josipkh (approved):

PR #58, opened 2025-04-03
Convenient access to wandb.init() kwargs

Confectio:
Make whole functionality of wandb init accessible without introducing a new field in the config for every kwarg. 
Instead now one field in the config `wandb_init_kwargs` can be specified, which will be passed to wandb.init().


JasperHoffmann (approved):
Looks Good!

Issue #27, opened 2025-03-06
Checkpointing and loading


JasperHoffmann:
Currently the checkpointing code can be improved. Potential improvements:

1. A simple load function that uses only one path to a experiment folder.
2. Storing not the root trainer module but one level below to make it easier to store and load components.
3. Use compositional programming by introducing a Checkpointer class to reduce the redundancy in the code. This can be then subclassed to provide checkpointers for specific algorithms eg SACCheckpointer.
4. Add functionality in the run script to load a trainer from another folder. The idea would be that we continue the run in another folder for safety reasons. In that way, we do not overwrite files.

Confectio:
I think this can be closed?

PR #60, opened 2025-04-09
Checkpointing

JasperHoffmann:
Based on #27 we updated the checkpointing capabilities of leap-c:

1. We can now continue training with the main run script.
2. The state dict of the trainer is saved in split modules, such that users can load only components of it to speed up evaluations workflows.
3. Also added git hash and git diff logging to main.
4. Furthermore, we added the resurrect module that allows to load cfgs, trainer and tasks from a ckpt dir.

We even added some tests!


leap_c/trainer.py
```diff
@@ -159,7 +159,7 @@ class Trainer(ABC, nn.Module):
     """
 
     def __init__(
-        self, task: Task, output_path: str | Path, device: str, cfg: BaseConfig
+        self, task: Task, output_path: str | Path | None, device: str, cfg: BaseConfig
```
Confectio, line 162: Specify in docstrings what happens if output_path is None?

leap_c/trainer.py
```diff
@@ -185,23 +188,24 @@ def __init__(
...
-        if cfg.log.wandb_logger:
-            if not cfg.log.wandb_init_kwargs.get("dir", False): #type:ignore               
-                wandbdir = self.output_path / "wandb"
-                wandbdir.mkdir(exist_ok=True)
-                cfg.log.wandb_init_kwargs["dir"] = wandbdir
-            wandb.init(
-                **cfg.log.wandb_init_kwargs   
-            )
+        if self.output_path is not None:
+            # init wandb
```
Confectio, line 192: I wonder if we should really hide the wandb logging behind output_path is not None. wandb.init() has its own functionality implemented that allows different modes of resuming runs, which we allow to be accessed by the wandb_init_kwargs: https://docs.wandb.ai/guides/runs/resuming/

leap_c/trainer.py
```diff
@@ -399,37 +406,86 @@ def _ckpt_path(self, name: str, suffix: str, basedir: Path | None = None) -> Pat
...
                 for i, opt in enumerate(self.optimizers)
             }
-            torch.save(state_dict, self._ckpt_path("optimizers", "pth"))
+            torch.save(state_dict, self._ckpt_path("optimizers", "ckpt", path))
 
-    def load(self, path: str | Path) -> None:
-        """Loads the state of a trainer from the output_path."""
-        # TODO (Jasper): Think about partial loading...
-        basedir = Path(path) 
+    def load(self, path: Path) -> None:
```
Confectio, line 457: Why not allow str, when we convert it to path in line 463 anyways? Also we allow str in the init for output path? (Im not for or against allowing string, but I think it should be kept consistent)

leap_c/trainer.py
```diff
@@ -399,37 +406,86 @@ def _ckpt_path(self, name: str, suffix: str, basedir: Path | None = None) -> Pat
...
 
         if self.cfg.val.ckpt_modus == "best":
             return basedir / "ckpts" / f"best_{name}.{suffix}"
         elif self.cfg.val.ckpt_modus == "last":
             return basedir / "ckpts" / f"last_{name}.{suffix}"
 
-        return basedir / "ckpts" / f"{self.step}_{name}.{suffix}"
+        return basedir / "ckpts" / f"{self.state.step}_{name}.{suffix}"
+
+    def save(self, path: Path | None = None) -> None:
```
Confectio, line 421: Allow string?(see comment below)

Confectio (changes requested):
Great PR :) . Only a few small comments

leap_c/trainer.py
JasperHoffmann: Yes!

leap_c/trainer.py
JasperHoffmann: However, we would need to assign a run id then :)

JasperHoffmann: Will try to update it accordingly.

JasperHoffmann: I think we should handle this in a separate PR :) For now it should be fine to just load it in a different wandb instance.

Confectio: I think we don't need to assign anything, since the user can set the wandb init options (including run id etc.). The user would have to do that manually then, but it is possible :)

Confectio (approved):

PR #61, opened 2025-04-11
Make cartpole balancing start randomly

Confectio:
Previously the initial state in the cartpole balancing started at 0, now the starting position is chosen randomly (Customizable via reset options).


JasperHoffmann (approved):
Looks good!

PR #62, opened 2025-04-11
Add MinMaxScalingExtractor

JasperHoffmann:
Added min-max-scaling into the extractor such that observations are now normalized. This can be useful for a lot of different  environments:

- Renamed IdentityExtractor to MinMaxScalingExtractor.
- Added a check to the min-max-scaling to ensure that there are correctly set min-max-bounds


Confectio (approved):
Looks good :)

PR #65, opened 2025-04-14
Proximal Policy Optimization (PPO) Trainer

MazenAmria:


leap_c/examples/cartpole/task.py
JasperHoffmann, line 1: We already have this task with pendulum_on_a_cart

JasperHoffmann: Maybe you can use this instead!

leap_c/rollout.py
JasperHoffmann, line 1: I think maybe you can check your stats again and directly make it a float, to keep it consistent!

leap_c/rl/ppo.py
```diff
@@ -0,0 +1,274 @@
...
+
+class PpoActor(nn.Module):
+    scale: torch.Tensor
+    loc: torch.Tensor
+
+    def __init__(self, task, env, mlp_cfg: MlpConfig):
+        super().__init__()
+
+        self.extractor = task.create_extractor(env)
+        action_dim = env.action_space.n
```
JasperHoffmann, line 90: For now we most likely want a continuous action space! So maybe you can use the squashed gaussian here.

leap_c/rl/ppo.py
```diff
@@ -0,0 +1,274 @@
...
+        self.mlp = MLP(
+            input_sizes=self.extractor.output_size,
+            output_sizes=int(action_dim),
+            mlp_cfg=mlp_cfg,
+        )
+
+    def forward(self, x: torch.Tensor, deterministic: bool = False, action=None):
+        e = self.extractor(x)
+        logits = self.mlp(e)
+        probs = Categorical(logits=logits)
```
JasperHoffmann, line 101: The same as above!

leap_c/rl/ppo.py
```diff
@@ -0,0 +1,274 @@
...
+            obs_collate = self.task.collate([obs], self.device)
+            with torch.no_grad():
+                value = self.q(obs_collate).cpu().numpy()[0]
+            action, log_prob, stats = self.act(obs)
+            self.report_stats("train_trajectory", {"action": action}, self.state.step)
+
+            obs_prime, reward, is_terminated, is_truncated, info = self.train_env.step(
+                action
+            )
+            done = float(is_terminated or is_truncated)
```
JasperHoffmann, line 188: I think is_terminated and is_truncated should be handled differently for learning the value function. Maybe we can just put both into the rollout buffer!

leap_c/rl/rollout_buffer.py
JasperHoffmann, line 1: Why can we not use the normal replay buffer?

JasperHoffmann (commented):
Going in the right direction. I added some comments :)

leap_c/rl/rollout_buffer.py
MazenAmria: It's possible to use the replay buffer, but in PPO, we want to collect the same number of steps and then discard them after updating the parameters. Using the replay buffer (a dequeue) for that didn't sound efficient, I'd keep pushing and popping out of it. Instead, I used a preallocated array that I overwrote after each update.

leap_c/trainer.py
```diff
@@ -322,6 +322,12 @@ def run(self) -> float:
         """Call this function in your script to start the training loop."""
         self.to(self.device)
 
+        for optimizer in self.optimizers:
+            for state in optimizer.state.values():
+                for k, v in state.items():
+                    if isinstance(v, torch.Tensor):
+                        state[k] = v.to(self.device)
+
```
MazenAmria, line 330: @JasperHoffmann I noticed a problem when I was continuing the training, basically loading the model and the optimizers _then_ moving the model to the device causes the optimizers' parameters to stay on CPU. This is just a temporary workaround.

PR #66, opened 2025-04-14
Use `set_flat` for speedup

Confectio:
Using acados' `set` methods for setting stagewise parameters caused a lot of overhead since the solver is reset to defaults after each `__call__` . Using the batched / unbatched `set_flat`, see [#1349](https://github.com/acados/acados/commit/f9af718b7c4bdde7b316091e5e9bb68b1071bea9), instead speeds this up a lot. Profiling an example shows total (!) training speedup of 20%!

Also updates acados to fix typo, see [#1500](https://github.com/acados/acados/pull/1500)


Confectio:
Sorry, I forgot to turn off the formatter, the relevant stuff is actually only until line 196...

FreyJo (approved):

PR #68, opened 2025-04-16
General improvements

Confectio:
- Use perf_counter to measure the time the whole batch solve takes and add it to the solver stats.
- set `qp_tol=1e-7` for the chain and the cartpole mpc, getting rid of a certain type of non-convergence (see comment below).
- update a docstring
- Fix bug in wandb logging
- Fix important bug concerning sensitivities
- Fix small bug in batch collate of batchiterates
- Change default Extractor in Task back to IdentityExtractor for now, because min max scaling can make problems with some observation spaces (see cartpole)

### DIFF
--- a/leap_c/examples/chain/mpc.py
+++ b/leap_c/examples/chain/mpc.py
@@ -1,4 +1,3 @@
-from copy import deepcopy
 from dataclasses import fields
 from pathlib import Path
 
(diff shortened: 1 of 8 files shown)


leap_c/examples/chain/mpc.py
JasperHoffmann, line 1: Maybe a short explanation why you set it that way :)

JasperHoffmann (approved):
Looks good!

leap_c/examples/chain/mpc.py
Confectio: It gets rid of a certain type of non-convergence (sometimes status 2 happens with a veeery low residual, simply because the qp solve is not accurate enough). This can be fixed by setting the qp tolerance slightly lower than the nlp tolerance (default is 1e-6, we set it to 1e-7 now). For the chain example this means getting rid of almost all of the non-convergences happening (depending on the initialization strategy, previously up to a few hundred for a 100k run, with qp_tol=1e-7, its not even 10 non-convergences in a 100k run).

JasperHoffmann (approved):
Looks good!

PR #69, opened 2025-04-17
long horizon swingup task

Confectio:
Avoid confusion in the future by creating a new task for cartpole swingup with the longer horizon


JasperHoffmann (approved):
Looks good!

PR #70, opened 2025-04-18
update constraints cartpole

Confectio:
Add terminal constraints to the cartpole formulation and make lbx and ubx reflect the actual threshold of the env.


JasperHoffmann (approved):
Looks good!

PR #72, opened 2025-04-23
Add test options and fixed objective

JasperHoffmann:
This PR:
- Adds a test cfg overwrite that can be used for doing integration tests. 
- We also fixed a bug tracking the best performance.
- Added the option to save no checkpoints.


dirkpr (commented):
Good to have a visible test config. As discussed, let's maximize scores to be aligned with accumulated rewards.

dirkpr (approved):

PR #74, opened 2025-04-29
Entropy scaling

JasperHoffmann:
This PR adds finer control to the entropy scaling and also bundles the configs of the SAC algorithms into one config:

- Add the option to manually set the target entropy.
- Add the option to set the lr_alpha to None, to deactivate autotuning of alpha.
- Removed SACFopConfigs and SACZopConfigs to avoid redundancy.


dirkpr (approved):
Looks good! Some notable code duplication due to the aim of single-file implementations for each algorithm.

PR #75, opened 2025-04-29
Clean point mass

JasperHoffmann:
This PR provides an update to the point mass environment:
- We cleaned up the old tasks to a single environment, where we need to navigate through a windfield.
- We removed the quadrotor script files and added a general one under a clean folder.


tests/leap_c/test_mpc.py
```diff
@@ -150,13 +150,13 @@ def test_backup_fn(learnable_point_mass_mpc_different_params: Mpc, n_batch: int)
         f"This test assumed state would be of type AcadosOcpFlattenedBatchIterate, but got {type(template_state)}"
     )
     ridiculous_state = AcadosOcpFlattenedBatchIterate(
```
dirkpr: Rename to nan_state

dirkpr (approved):

PR #76, opened 2025-04-30
Fix inheritance in Cart Pole

Confectio:
This fixes the Cart Pole task with long horizon accidentally inheriting from Task instead of from the other swingup task (and hence, crashing due to abstract methods).


dirkpr (approved):
Added a minor todo for refactor later.

PR #78, opened 2025-05-07
Cartpole params

josipkh:
Found two small discrepancies in the cartpole example, not sure if they were intentional.
If yes, I would make them more explicit.
If not, the fix should improve the results :)

### DIFF
--- a/leap_c/examples/pendulum_on_a_cart/env.py
+++ b/leap_c/examples/pendulum_on_a_cart/env.py
@@ -27,7 +27,7 @@ class PendulumOnCartSwingupEnv(gym.Env):
     the possible range of the cart's center of mass in the observation space,
     but the episode terminates if it leaves the interval (-2.4, 2.4) already.
     NOTE: The pole angle is actually bounded between -2pi and 2pi by always adding/subtracting
-    (in the negative / in the positive case) the highest multiple of 2.4
+    (in the negative / in the positive case) the highest multiple of 2pi
     until the pole angle is within the bounds again.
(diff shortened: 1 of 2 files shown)


Confectio:
Good catch, to be honest, the lenght of the pole was not intentional :') . 
I also noticed the weird length of the horizon at some point (6 instead of 5), but since the MPC with fixed reference angle 0 was not able to perform sufficiently good already, I thought it did not matter and left it like this.
I think it is actually nice to have a mismatch between the MPC dynamics and the environment dynamics.
But not sure if we should rather have this in an extra task with a more meaningful mismatch instead. 
Maybe also with a mismatch in the simulation time of a step (like for N=6 and T=0.25s, a "step" in the MPC simulates roughly 0.04 seconds, but a step in the environment simulates 0.05s)?

Did you test training Cart Pole with the new parameters by chance?

JasperHoffmann:
I think we should improve the parameter management here, so there is a single source. We can discuss in more detail for the refactor!

josipkh:
Thanks for the clarification :)
I agree that some mismatch should be used, probably it's best to have a separate task or models of varying complexity (as in the quadrotor example).
On the other hand, I don't think that different sampling times for the controller and the simulation are the best way to go. Perhaps a more realistic scenario is to add some delay to the observations or before applying actions.

I didn't do a full run, but a quick comparison (50k samples) doesn't show a big difference in learning with the update, at least from my understanding.

I would suggest to merge this as it is now and do a proper refactoring a bit later (I can open a new issue).

dirkpr (approved):
Small change. More in the refactor.

PR #79, opened 2025-05-09
Using `TransformedDistribution` for `SquashedGaussian` implementation

MazenAmria:


JasperHoffmann:
Looks good, I think we should do a benchmark though. The last time I used this our performance dropped.

PR #80, opened 2025-05-14
Logger

JasperHoffmann:
This PR refactors the logging in leap-c, to make it more consistent and easier to maintain in the future:

1. A new class `Logger`, to remove logging functionality from the trainer:
     - Stats can be report aperiodically and in groups.
2. Smoothed logging:
     - Per default a statistic is smoothed over `10.000` steps and is reported every `1.000` steps:
     - This drastically reduces the number of stats logged and reduces the variance of the metrics.
     - You can turn off this feature by a `with_smoothing` flag, when reporting a stat. This is used for example for the validation results.
     - To have a simple configuration, the smoothing window and the report interval of smoothed statistics can be only changed globally.
3. Verbosity levels:
     - We also added a verbose flag, which can be chosen when reporting the stats.
     - When the logger is not in verbose mode it ignores all verbose stats, to reduce the number of stats tracked. 
     - The flag can be set in the main run script and per default verbose logging is off.


leap_c/logger.py
```diff
@@ -0,0 +1,250 @@
...
+        if prev_timestamp == -1:
+            return None
+
+        interval_idx = (timestamp + 1) // self._interval
+        interval_idx_prev = (prev_timestamp + 1) // self._interval
+
+        if interval_idx == interval_idx_prev:
+            return None
+
+        def clean_until(t: int) -> None:
```
dirkpr, line 96: Consider moving this out of/ parallel to update

leap_c/logger.py
```diff
@@ -0,0 +1,250 @@
...
+            interval: The interval at which statistics will be logged.
+            window_size: The moving window size for the statistics.
+        """
+        self._interval = interval
+        self._window_size = window_size
+
+        self._timestamps: dict = defaultdict(list)
+        self._statistics: dict = defaultdict(list)
+
+    def update(
```
dirkpr, line 59: Functionality looks good! But update appears like we should make it clearer by refactoring its content into smaller subfunctions

leap_c/rl/sac.py
```diff
@@ -74,7 +75,7 @@ class SacBaseConfig(BaseConfig):
     sac: SacAlgorithmConfig = field(default_factory=SacAlgorithmConfig)
     train: TrainConfig = field(default_factory=TrainConfig)
     val: ValConfig = field(default_factory=ValConfig)
-    log: LogConfig = field(default_factory=LogConfig)
+    log: LoggerConfig = field(default_factory=LoggerConfig)
```
dirkpr, line 78: Rename to logger for consistent variable names

dirkpr (approved):
Looks good! More in the refactor to come!

PR #82, opened 2025-05-16
Structure refactor

JasperHoffmann:
This PR updates the structure of the code:

- Acados becomes its own submodule to start decoupling from the Acados code. Thus, the acados pytorch layer is also moved into that subfolder.
- We provide a new way how to print dataclass configs
- Tests are now properly split according to the submodules.
- We split the different utility functions into separate modules.


dirkpr (approved):
New structure looks good! Should restructure in a next PR to move acados into an "ocp" folder at the same level as rl and nn

PR #83, opened 2025-05-21
Add torch autograd functionality.

JasperHoffmann:
This PR prepares the acados module refactor by providing a generic way to create PyTorch autograd Functions.

**Motivation:** Decoupling the creation of autograd functionalities for PyTorch from the acados sensitivity calculations, allows wrapping the differentiable capabilities also for Jax at a later point.

**Benefits**:
- Improving our current implementation of the PyTorch layer for acados, by being more general.
- We introduce a new `custom_ctx` object for layers that can be wrapped. This results in an explicit handling of the information shared between the forward and backward pass in leap_c. In some cases, we want to do more than just forward and backward passes but also calculate specific sensitivities. The ctx object outside of PyTorch allows then to store the information even longer then in a traditional forward and backward scheme.

### DIFF
--- a/leap_c/utils/torch_autograd.py
+++ b/leap_c/utils/torch_autograd.py
@@ -20,7 +20,7 @@ def _add_none(data):
 
 def create_autograd_function(obj) -> type[torch.autograd.Function]:
     """
-    Create a PyTorch autograd function from an object implementing
+    Creates a PyTorch autograd function from an object implementing
     NumPy-based forward and backward methods.
 
     An example:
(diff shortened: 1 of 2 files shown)


leap_c/utils/torch_autograd.py
```diff
@@ -0,0 +1,77 @@
...
+                # Compute gradients using NumPy arrays
+                return grad_inputs  # A single gradient or a tuple
+
+    Args:
+        obj: An object implementing `forward(custom_ctx, *args)` and
+             `backward(custom_ctx, *grad_outputs)` using NumPy.
+
+    Returns:
+        A callable function that behaves like a PyTorch autograd
+        function.
```
FreyJo, line 46: maybe "behaves like" is a bit weak, as it actually is an `autograd.Function` via inheritance.

FreyJo (approved):
Looks good to me!

PR #84, opened 2025-05-26
Update folder structure

JasperHoffmann:
This PR updates the folder structure to allow a cleaner separation of functionality:
1. We introduce a `torch` folder, where RL implementations like the different SAC trainers are stored.
2. We provide an `ocp` folder, where interfaces to OCP solvers like Acados are stored.
3. The `autograd` folder contains all the code necessary for wrapping code such that it can be used for `backward` and `forward` passes in Jax or PyTorch.

We update the test folder structure accordingly.

### DIFF
--- a/leap_c/examples/chain/mpc.py
+++ b/leap_c/examples/chain/mpc.py
@@ -124,7 +124,7 @@ def export_parametric_ocp(
     fix_point: np.ndarray = np.array([0.0, 0.0, 0.0]),
     pos_last_mass_ref: np.ndarray = np.array([1.0, 0.0, 0.0]),
 ) -> tuple[AcadosOcp, DMStruct]:
-    # create fun object to formulate the OCP
+    # create ocp object to formulate the OCP
     ocp = AcadosOcp()
     ocp.solver_options.N_horizon = N_horizon
     ocp.solver_options.Tsim = tf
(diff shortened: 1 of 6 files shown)


leap_c/examples/chain/mpc.py
```diff
@@ -124,7 +124,7 @@ def export_parametric_ocp(
     fix_point: np.ndarray = np.array([0.0, 0.0, 0.0]),
     pos_last_mass_ref: np.ndarray = np.array([1.0, 0.0, 0.0]),
 ) -> tuple[AcadosOcp, DMStruct]:
-    # create ocp object to formulate the OCP
+    # create fun object to formulate the OCP
```
JasperHoffmann, line 127: This should be not renamed!

leap_c/examples/mujoco/reacher/mpc.py
```diff
@@ -293,7 +293,7 @@ def export_parametric_ocp(
 
 def configure_ocp_solver(ocp: AcadosOcp, exact_hess_dyn: bool):
     ocp.solver_options.integrator_type = "DISCRETE"
-    # ocp.solver_options.nlp_solver_type = "SQP_RTI"
+    # fun.solver_options.nlp_solver_type = "SQP_RTI"
```
JasperHoffmann, line 296: This should have not been renamed

dirkpr (approved):
Looks good!

PR #88, opened 2025-06-04
Acados Implicit Function

JasperHoffmann:
This PR improves the structure of the Acados module:

1. We split the MPC class into different classes to reduce the complexity of the code.
2. We introduce the `AcadosImplicitFunction` which provides a `forward` and `backward` pass that can be used to integrate it into Jax or PyTorch.
3. We create an explicit `AcadosInitializerClass` that provides the logic how to initialize a solver.
4. We further provide an `AcadosImplicitModule` which is the central point for interacting with Acados for implicit differentiation in PyTorch.


leap_c/ocp/acados/implicit.py
```diff
@@ -0,0 +1,234 @@
...
+            grad_p_global = _safe_sum(
+                _back(value_grad, "dvalue_dp_global"),
+                _back(u0_grad, "du0_dp_global"),
+                _back(x_grad, "dx_dp_global"),
+                _back(u_grad, "du_dp_global"),
+            )
+        else:
+            grad_p_global = None
+
+        return (None, None, grad_u0, grad_p_global, None, None, None)
```
FreyJo, line 190: How is inferred/specified what the outputs of backwards are?

leap_c/ocp/acados/initializer.py
```diff
@@ -0,0 +1,87 @@
...
+from acados_template.acados_ocp_solver import AcadosOcpSolver
+import numpy as np
+
+from leap_c.ocp.acados.data import (
+    AcadosSolverInput,
+    _collate_acados_flattened_iterate_fn,
+)
+
+
+class AcadosInitializer(ABC):
```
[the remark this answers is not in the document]
Confectio: Not counting more involved manipulation (like training), all initialization strategies I implemented so far could lastly be interfaced by only only using `mpc_input` and the previous iterate as input for the function call immediately creating the initialization iterate (to be precise, I mostly used line 113 in sac_fop.py by setting `self.prepare_mpc_state`). So with adding the previous solution like @FreyJo suggested, this signature should cover a good amount of initialization strategies.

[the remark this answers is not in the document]
JasperHoffmann: I think it is a little bit tricky here, as the initialization strategies that @Confectio implemented already work from the RL side. As we first load the iterate that comes from the `ctx` object and use the `AcadosInitializer` only for retries or when the `ctx` is None, which happens at the start of episodes when generating new data.

In general, the question is what belongs into `acados` module and what belongs into the rl part. Will think about one more time :) But any suggestions are welcome!

leap_c/ocp/acados/utils/create_solver.py
```diff
@@ -0,0 +1,328 @@
...
+    if ocp.cost.cost_type not in ["EXTERNAL"]:
+        ocp.translate_intermediate_cost_term_to_external(
+            cost_hessian=ocp.solver_options.hessian_approx
+        )
+    if ocp.cost.cost_type_e not in ["EXTERNAL"]:
+        ocp.translate_terminal_cost_term_to_external(
+            cost_hessian=ocp.solver_options.hessian_approx
+        )
+
+    _turn_on_warmstart(ocp)
```
Confectio, line 58: Not sure if we should add this by default. As far as I can see, this (still) only concerns the warmstarting of the first QP being solved.
We added this in the context of warmstarting for LMPC and we also left this in for the experiments, but I didn't check how much of a difference this makes for warmstarting NLPs. Can you add a TODO? I can check this when Im back

leap_c/ocp/acados/utils/create_solver.py
```diff
@@ -0,0 +1,328 @@
...
+
+    if add_delete_hook:
+        DeleteDirectoryHook(batch_solver, export_directory)
+
+    return batch_solver
+
+
+def create_forward_backward_batch_solvers(
+    ocp: AcadosOcp,
+    sensitivity_ocp: AcadosOcp | None = None,  # type:ignore
```
Confectio, line 99: Should this type:ignore still be there?

leap_c/ocp/acados/utils/create_solver.py
```diff
@@ -0,0 +1,328 @@
...
+    export_directory: str | Path | None = None,
+    discount_factor: float | None = None,
+    n_batch_max: int = 256,
+    num_threads: int = 4,
+):
+    # check if we can use the forward solver for the backward pass.
+    need_backward_solver = _check_need_sensitivity_solver(ocp)
+
+    if need_backward_solver:
```
Confectio, line 108: should be `if not need_backward_solver`?

leap_c/ocp/acados/utils/create_solver.py
```diff
@@ -0,0 +1,328 @@
...
+        discount_factor=discount_factor,
+        n_batch_max=n_batch_max,
+        num_threads=num_threads,
+    )
+
+    if not need_backward_solver:
+        return forward_batch_solver, forward_batch_solver
+
+    if sensitivity_ocp is None:
+        sensitivity_ocp: AcadosOcp = deepcopy(
```
Confectio, line 124: Probably merge this (and following lines) with `derive_sensitivity_ocp` (there are some settings that are being set which are not being set in `derive_sensitivty_ocp`) and then use `derive_sensitivity_ocp`?

leap_c/ocp/acados/utils/prepare_solver.py
```diff
@@ -0,0 +1,192 @@
...
+        param_default = ocp.model.p.tile(batch_size, N + 1)  # type:ignore
+        param = param_default.reshape(batch_size, -1)
+        batch_solver.set_flat("p", param)
+    elif p_stagewise is not None and p_stagewise_sparse_idx is None:
+        # if p_stagewise is provided, set it
+        param = p_stagewise.reshape(batch_size, -1)
+        batch_solver.set_flat("p", param)
+    elif p_stagewise is not None and p_stagewise_sparse_idx is not None:
+        # if p_stagewise is provided and sparse indices are provided, set it
+        for idx, stage in zip(range(batch_size), range(N + 1)):
```
Confectio, line 149: Bug? (This only sets stage 1 for solver 1, stage 2 for solver 2, etc.?)
Maybe we should add a test for stagewise params.

leap_c/ocp/acados/utils/solve.py
```diff
@@ -0,0 +1,70 @@
...
+    time_solve = time.perf_counter() - start
+
+    active_solvers = batch_solver.ocp_solvers[:batch_size]
+    batch_status = np.array([solver.status for solver in active_solvers])
+
+    if with_retry and any(status != 0 for status in batch_status):
+
+        for idx, solver in enumerate(active_solvers):
+            single_iterate = initializer.single_iterate(solver_input.get_sample(idx))
+            solver.load_iterate_from_flat_obj(single_iterate)
```
Confectio, line 56: Only retry iterates that did not converge (I mean add additional status != 0 check before setting an iterate from the initializer)

leap_c/ocp/acados/utils/solve.py
```diff
@@ -0,0 +1,70 @@
...
+
+        start_retry = time.perf_counter()
+        batch_solver.solve(n_batch=batch_size)
+        time_solve += time.perf_counter() - start_retry
+
+    batch_status_retry = np.array([solver.status for solver in active_solvers])
+
+    stats = {
+        "solving_time": time_solve,
+        "success_rate": batch_status_retry.mean(),
```
Confectio, line 66: I think this is a bug (Shouldnt call mean() over the status, but over the booleans of status == 0)

tests/leap_c/ocp/acados/test_acados_torch.py
```diff
@@ -0,0 +1,502 @@
...
+        #     test_data.p_global,
+        #     "fine_eps",
+        # ),
+        # (
+        #     "dQ/du0",
+        #     _create_dQdu0_test(implicit_layer, test_data.x0, test_data.p_global),
+        #     test_data.u0,
+        #     "standard",
+        # ),
+    ]
```
Confectio, line 485: Should this and the above be commented-out?

leap_c/ocp/acados/layer.py
```diff
@@ -1,3 +1,4 @@
+# OUTDATED: This file is deprecated and will be removed in a couple of PRs.
```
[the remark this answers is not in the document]
JasperHoffmann: We use now the check from the following PR in acados: https://github.com/acados/acados/pull/1548

leap_c/ocp/acados/implicit.py
```diff
@@ -0,0 +1,317 @@
...
+            x0: Initial states with shape `(B, x_dim)`.
+            u0: Initial actions with shape `(B, u_dim)`. Defaults to `None`.
+            p_global: Global parameters shared across all stages,
+                shape `(B, p_global_dim)`. Defaults to `None`.
+            p_stagewise: Stagewise parameters.
+                If `p_stagewise_sparse_idx` is `None`, shape is
+                `(B, N+1, p_stagewise_dim)`.
+                If `p_stagewise_sparse_idx` is provided, shape is
+                `(B, N+1, len(p_stagewise_sparse_idx))`.
+                For multi-phase MPC with `p_stagewise_sparse_idx`, this is
```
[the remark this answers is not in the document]
JasperHoffmann: Yes, updated the docstring.

leap_c/ocp/acados/implicit.py
```diff
@@ -0,0 +1,317 @@
...
+        # fetch output
+        active_solvers = self.forward_batch_solver.ocp_solvers[:batch_size]
+        sol_iterate = self.forward_batch_solver.store_iterate_to_flat_obj(
+            n_batch=batch_size
+        )
+        ctx = AcadosImplicitCtx(
+            iterate=sol_iterate, log=log, status=status, solver_input=solver_input
+        )
+        sol_value = np.array([s.get_cost() for s in active_solvers])
+        sol_u0 = sol_iterate.u[:, : self.ocp.dims.nu]
```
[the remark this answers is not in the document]
JasperHoffmann: Soon :D

leap_c/ocp/acados/utils/create_solver.py
JasperHoffmann: Sounds good!

leap_c/ocp/acados/utils/prepare_solver.py
JasperHoffmann: Good catch, changed it to a product.

leap_c/ocp/acados/utils/prepare_solver.py
```diff
@@ -0,0 +1,192 @@
...
+    # initial conditions
+    for idx, solver in enumerate(active_solvers):
+        solver.set(0, "x", x0[idx])
+        solver.constraints_set(0, "lbx", x0[idx])
+        solver.constraints_set(0, "ubx", x0[idx])
+
+        if u0 is not None:
+            solver.set(0, "u", u0[idx])
+            solver.constraints_set(0, "lbu", u0[idx])
+            solver.constraints_set(0, "ubu", u0[idx])
```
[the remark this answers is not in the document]
JasperHoffmann: Added defaults!

dirkpr (approved):

PR #99, opened 2025-07-22
Added load codegen option

JasperHoffmann:
This PR allows to load generated code that was compiled in previous runs:
- `run_sac_fop.py` and `run_sac_zop.py` get an additional `reuse_code` flag, that stores and loads compiled code in `output/controller_code`
- Additionally, we fixed a small inconsistency, that `SacZopTrainer` had a different default extractor.

### DIFF
--- a/scripts/run_sac_fop.py
+++ b/scripts/run_sac_fop.py
@@ -19,24 +19,14 @@ class RunSacFopConfig:
 
 
 def run_sac_fop(
-    output_path: str | Path,
+    trainer_output_path: str | Path,
     env_name: str,
     controller_name: str,
     seed: int = 0,
     device: str = "cuda",
     verbose: bool = True,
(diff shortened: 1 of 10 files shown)


scripts/run_sac_fop.py
```diff
@@ -100,19 +101,35 @@ def run_sac_fop(
...
 
+    if args.output_path is None:
+        trainer_output_path = default_output_path(
+            seed=args.seed, tags=["sac_fop", args.env, args.controller]
+        )
+        reuse_code_dir = (
+            default_controller_code_path() if args.reuse_code else None
+        )
+    else:
+        assert not args.reuse_code, "Cannot reuse code when output_path is specified."
```
Confectio, line 123: Why not?

Confectio (approved):
👍

PR #100, opened 2025-07-25
Logging: minor fixes

josipkh:
This PR proposes the following changes:
1. fix accessing `wandb_init_kwargs`, e.g., when setting custom run names
2. add an initial policy validation (perhaps the initial validation score does not start at zero)
3. move closing of the Logger before the `return` command, otherwise it does not close after the run in some cases

Also, I noticed that when `val_report_score` is `"final"`, the best score is returned instead of the last one.
Should a `"best"` option be added and the rest modified accordingly?

### DIFF
--- a/leap_c/trainer.py
+++ b/leap_c/trainer.py
@@ -198,13 +198,17 @@ class Trainer(ABC, nn.Module, Generic[TrainerConfigType]):
 
         train_loop_iter = self.train_loop()
 
+        # initial policy validation
+        val_score = self.validate()
+        self.state.scores.append(val_score)
+        self.state.max_score = val_score
 
         while self.state.step < self.cfg.train_steps:
             # train
             self.state.step += next(train_loop_iter)
 
             # validate
-            if self.state.step // self.cfg.val_interval > len(self.state.scores):
+            if self.state.step // self.cfg.val_interval > len(self.state.scores)-1:
                 val_score = self.validate()
                 self.state.scores.append(val_score)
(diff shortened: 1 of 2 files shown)


leap_c/trainer.py
```diff
@@ -198,13 +198,17 @@ def run(self) -> float:
...
+        self.state.scores.append(val_score)
+        self.state.max_score = val_score
 
         while self.state.step < self.cfg.train_steps:
             # train
             self.state.step += next(train_loop_iter)
 
             # validate
-            if self.state.step // self.cfg.val_interval > len(self.state.scores):
+            if self.state.step // self.cfg.val_interval > len(self.state.scores)-1:
```
JasperHoffmann, line 211: Could you reformat it to ` - 1` :)

leap_c/utils/logger.py
```diff
@@ -154,7 +154,7 @@ def __init__(self, cfg: LoggerConfig, output_path: str | Path) -> None:
 
         # init wandb
         if cfg.wandb_logger:
-            if not cfg.log.wandb_init_kwargs.get("dir", False):  # type:ignore
+            if not cfg.wandb_init_kwargs.get("dir", False):  # type:ignore
```
JasperHoffmann, line 157: Thanks!

JasperHoffmann (approved):
Looks good!

JasperHoffmann (approved):
Looks good!

PR #101, opened 2025-08-04
Standardizes scripts and more

JasperHoffmann:
Hvac:
- Bugfixes 
- Makes it accessible via the run scripts.
- Adds noise to the dynamics (such that they are different from the MPC dynamics)
- Adds noise to the forecasts
- Adds dependency of the reward on the price data

General:
- Changes the shape of trajectory tensors returned by the MPC layer (from (Batch, Flat) to (Batch, Horizon, dim)
- Add tests for the sensitivities regrading the x trajectory and the u trajectory
- Fix rare bug where sensitivity computation would fail if ONLY adjoint sensitivities were needed in the backward
- Fix seeds for adjoint sensitivities of the x trajectory and the u trajectory
- Fix bug where some sensitivities would be computed unnecessarily, when dL/du would be zero, but dL/dx wasn't or vice versa.
- Bring back git repo logging
- Change default param interface to take observations into account
- Improved logging
- Improved scripts for running the MPC controller, SAC, SAC-FOP and SAC-ZOP


leap_c/run.py
```diff
@@ -48,5 +53,6 @@ def init_run(trainer: Trainer, cfg, output_path: str | Path):
         trainer.load(output_path)
 
     # store git hash and diff
-    log_git_hash_and_diff(output_path / "git.txt") 
+    module_root = Path(leap_c.__file__).parent.parent
```
Confectio, line 56: Need to investigate this, therefore leaving a comment here

leap_c/torch/rl/sac_fop.py
```diff
@@ -243,20 +246,19 @@ def train_loop(self) -> Iterator[int]:
                 action = pi_output.action.cpu().numpy()[0]
                 param = pi_output.param.cpu().numpy()[0]
 
-            self.report_stats(
-                "train_trajectory", {"param": param, "action": action}, verbose=True
-            )
+            # self.report_stats(
+            #     "train_trajectory", {"param": param, "action": action}, verbose=True
+            # )
```
Confectio, line 251: Uncomment?

scripts/run_controller.py
```diff
@@ -44,45 +45,42 @@ def __init__(
...
-        
+
     def train_loop(self) -> Iterator[int]:
         """No training - just return immediately."""
-        yield 1
-        
+        step = 0
+        while True:
+            yield step
+            step += 1
```
Confectio, line 57: Why this approach instead of always returning 1?

Confectio (commented):
Leaving this review for now to work out the changes that still need to be done

Confectio (approved):

Confectio:
Me approving a PR where I did some of the work 
image

PR #103, opened 2025-08-05
Introducing automated Ruff Linting and Formatting

Confectio:
- Introduces a github action that fails if the files are not linted and formatted according to the standard ruff linting and formatting. 

- Introduces a pre-commit configuration for running the ruff linting and formatting automatically before every commit

- Every python file in leap_c\ , tests\ and scripts\ has been linted and formatted.

### DIFF
--- a/leap_c/examples/cartpole/config.py
+++ b/leap_c/examples/cartpole/config.py
@@ -11,7 +11,7 @@ class CartPoleParams:
     M: Parameter  # mass of the cart [kg]
     m: Parameter  # mass of the ball [kg]
     g: Parameter  # gravity constant [m/s^2]
-    l: Parameter  # length of the rod [m]
+    l: Parameter  # length of the rod [m] # noqa
 
     # Cost matrix factorization parameters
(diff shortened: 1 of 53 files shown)


leap_c/examples/cartpole/config.py
```diff
@@ -11,7 +11,7 @@ class CartPoleParams:
     M: Parameter  # mass of the cart [kg]
     m: Parameter  # mass of the ball [kg]
     g: Parameter  # gravity constant [m/s^2]
-    l: Parameter  # length of the rod [m]
+    l: Parameter  # length of the rod [m] # noqa
```
dirkpr, line 14: I would suggest we add a ruff.toml where we can disable specified rules. We can also explicitly define formatting there.

dirkpr (changes requested):
Looks good in general! But I would suggest we add a `ruff.toml` to define linting and formatting explicitly. That allows us to share a config that is not the default and specify rules to enable or disable. Same with formatting specs.

dirkpr (approved):

PR #104, opened 2025-08-11
Trainer: return value of `run`

josipkh:
I would suggest to slightly update the return value of `Trainer.run()`.
Also, there was a warning with the chain example related to [this thread](https://stackoverflow.com/questions/60149105/userwarning-warn-box-bound-precision-lowered-by-casting-to-float32).


josipkh:
Thank you for the suggestion :)

PR #107, opened 2025-08-19
Providing `env`, `controller`, and `seed` to `create_cfg` to dynamically configure the run based on these arguments.

MazenAmria:


leap_c/ocp/acados/data.py
```diff
@@ -67,13 +67,13 @@ def collate_acados_flattened_batch_iterate_fn(
...
-        su=np.concat([x.su for x in batch], axis=0),
-        pi=np.concat([x.pi for x in batch], axis=0),
-        lam=np.concat([x.lam for x in batch], axis=0),
+        x=np.concatenate([x.x for x in batch], axis=0),
+        u=np.concatenate([x.u for x in batch], axis=0),
+        z=np.concatenate([x.z for x in batch], axis=0),
+        sl=np.concatenate([x.sl for x in batch], axis=0),
+        su=np.concatenate([x.su for x in batch], axis=0),
+        pi=np.concatenate([x.pi for x in batch], axis=0),
+        lam=np.concatenate([x.lam for x in batch], axis=0),
```
MazenAmria, line 76: this is unrelated to the PR, but it was failing when running sac fop on cpu because there's no `np.concat` but `np.concatenate`, so I fixed it here.

JasperHoffmann (changes requested):
Looks great, I would suggest only a small change :)

PR #109, opened 2025-08-27
Const param mlp

JasperHoffmann:
Provides a simple way to replace the `Mlp` block with a constant parameter. Setting `hidden_dims` of the `MlpCfg` to either `None` or an empty sequence will replace the MLP with a constant parameter.


Confectio (approved):
Nice :)

PR #113, opened 2025-09-03
Fix linting, update acados, small bugfix

Confectio:
Supposedly fixes that linting is run on PRs from forks now

Also comes with an acados update and a small bugfix which would trigger an error when trying to use levenberg marquardt regularization


dirkpr (approved):
Looks good!

PR #115, opened 2025-09-05
Changes to ruff

FilippoAiraldi:
In the wake of [leap-c#112](https://github.com/leap-c/leap-c/pull/112), this PR merges `ruff.toml` into `pyproject.toml` and revises some of its config.

In more details, the first commit

- removes Ruff from the core dependencies of leap-c
- unifies `ruff.toml` with `pyproject.toml`, as the latter is meant to contain as many settings/configurations for the project as possible
- adds `target-version = "py311"` in accordance with leap-c's Python requirement
- switches from `exclude` to `extend-exclude`; in this way, predefined excluded directories (such as `".venv"`) are not overwritten
- adds fix also for isort (see `extend-select` with option `I`)

The second commit includes the fixes after running `ruff check --fix` with the new configuration.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ files: '^(tests|leap_c|scripts)(/|\\).*\.py$'
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.12.7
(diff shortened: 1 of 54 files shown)


JasperHoffmann (approved):
Looks good!

PR #116, opened 2025-09-05
Logger refactoring

FilippoAiraldi:
Refactoring of the logger class (in `leap_c/utils/logger.py`) to improve efficiency and to remove unnecessary (optional) packages

In particular, [commit 26abb45](https://github.com/FilippoAiraldi/leap-c/commit/26abb45aecd85dd5fb405e10a8429b92067cd871) 
- replaces lists with `deque`, allowing for a much faster deletion of elements in the front (`deque.popleft` runs in `O(1)`, while `del list[0]` in `O(n)`).
- unifies timestamps and statistics in a single `deque` of 2-tuples, so that we only have to pop from one `deque` instead of two
- introduces an additional running sum for each statistics. At the cost of a bit more memory, keeping a running sum allows for faster computation of the average over a window without the need of iterating over said window.

[Commit 55d03d5](https://github.com/FilippoAiraldi/leap-c/commit/55d03d59b74830f8d9cabd5e611bf71e6532aa4e) moves imports for `wandb` and `torch.utils.tensorboard` inside their respective conditional blocks/functions. Previously, even if not activated (e.g., only logging to CSV), these would be imported and could result in an ImportError if the user did not install them.

### DIFF
--- a/leap_c/utils/logger.py
+++ b/leap_c/utils/logger.py
@@ -1,5 +1,5 @@
 import bisect
-from collections import defaultdict
+from collections import defaultdict, deque
 from collections.abc import Generator
 from dataclasses import dataclass, field
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


dirkpr:
Good stuff!

PR #117, opened 2025-09-05
Removed class attributes in leap_c/ocp/acados/parameters.py

FilippoAiraldi:
This PR removes (probably unintentional) class attributes from `AcadosParameter` and `AcadosParameterManager`. 

Class attributes are attributes defined at the class level and assigned a default value. These attributes are then **shared across all instances** of that class. From the code, this does not seem to the intended purpose, e.g., each `AcadosParameterManager` instance should have its own `parameters` dict.

I also turned `AcadosParameter` to a dataclass so as to allow `vary_stages` to have a default empty list value. In the current implementation, the default value is mutable, thus shared across all instances, as aforementioned. 

I also removed `None` as a possible value from `learnable_parameters` and similar fields in `AcadosParameterManager`. It seems these fields are always initialized at construction and are never nulled.

### DIFF
--- a/leap_c/ocp/acados/parameters.py
+++ b/leap_c/ocp/acados/parameters.py
@@ -1,4 +1,5 @@
-from typing import NamedTuple, Literal
+from dataclasses import dataclass, field
+from typing import Literal
 import warnings
 
 import casadi as ca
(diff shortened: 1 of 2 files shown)


JasperHoffmann:
Thank you, makes sense! The `AcadosParameterManager` should just have type hints on the class level :)

JasperHoffmann (approved):
Looks good!

PR #118, opened 2025-09-05
exposing `n_batch_max` and `num_threads_batch_solver` in `AcadosDiffMpc.__init__`

FilippoAiraldi:
Changing `N_BATCH_MAX` and `NUM_THREADS_BATCH_SOLVER` to different values (e.g., in my case to higher values for Monte Carlo computations) is not straightforward. Before importing any other leap-c module, one must

```python
import leap_c.ocp.acados.diff_mpc as leap_c_diff_mpc

leap_c_diff_mpc.N_BATCH_MAX = ...
leap_c_diff_mpc.NUM_THREADS_BATCH_SOLVER = ...

from leap_c import ...  # other leap-c imports
```
so that the new values for these two constants are actually used.

By instead exposing two new arguments in `AcadosDiffMpc.__init__` we could have them passed as arguments when creating the controller.

### DIFF
--- a/leap_c/ocp/acados/diff_mpc.py
+++ b/leap_c/ocp/acados/diff_mpc.py
@@ -93,6 +93,9 @@ class AcadosDiffMpcFunction(DiffFunction):
         sensitivity_ocp: AcadosOcp | None = None,
         discount_factor: float | None = None,
         export_directory: Path | None = None,
+        *,
+        n_batch_max: int = N_BATCH_MAX,
+        num_threads_batch_solver: int = NUM_THREADS_BATCH_SOLVER,
     ) -> None:
         self.ocp = ocp
         self.forward_batch_solver, self.backward_batch_solver = (
@@ -101,8 +104,8 @@ class AcadosDiffMpcFunction(DiffFunction):
                 sensitivity_ocp=sensitivity_ocp,
                 discount_factor=discount_factor,
                 export_directory=export_directory,
-                n_batch_max=N_BATCH_MAX,
-                num_threads=NUM_THREADS_BATCH_SOLVER,
+                n_batch_max=n_batch_max,
+                num_threads=num_threads_batch_solver,
             )
         )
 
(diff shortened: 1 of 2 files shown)


JasperHoffmann:
I think adding something like this makes sense :) I still think there is a value in having global variables. I think one problem could be that the variables are just accessed while parsing the class definition of `AcadosDiffMpc`. An alternative could be to allow as a default value `None` for both `n_batch_max` and `num_threads` for the arguments of `AcadosDiffMpc.__init__`. We then access the defaults then like this:

```python
        self.forward_batch_solver, self.backward_batch_solver = (
            create_forward_backward_batch_solvers(
                ocp=ocp,
                sensitivity_ocp=sensitivity_ocp,
                discount_factor=discount_factor,
                export_directory=export_directory,
                n_batch_max=N_BATCH_MAX if n_batch_max is None else n_batch_max,   # reevaluate what the globals are
                num_threads=NUM_THREADS_BATCH_SOLVER if num_threads is None else num_threads,
            )
        )
  ```

**Comment by FilippoAiraldi** [comment]
Hi @JasperHoffmann,  

Yeah, I don't see any issue with keeping the global variables defined at the top of the file. Maybe just change their names to something like, e.g., `DEFAULT_N_BATCH_MAX` and `DEFAULT_NUM_THREADS_BATCH_SOLVER`?

File: leap_c/ocp/acados/diff_mpc.py [hunk 1]
```diff
@@ -104,8 +104,10 @@ def __init__(
                 sensitivity_ocp=sensitivity_ocp,
                 discount_factor=discount_factor,
                 export_directory=export_directory,
-                n_batch_max=n_batch_max,
-                num_threads=num_threads_batch_solver,
+                n_batch_max=N_BATCH_MAX if n_batch_max is None else n_batch_max,
```
dirkpr, line 107: Renaming to `DEFAULT_N_BATCH_MAX`, `DEFAULT_NUM_THREADS_BATCH_SOLVER` was a good idea.

JasperHoffmann (approved):
Looks good :) Thank you!

Btw. is there a specific reason for making `n_batch_max`, and `num_threads_batch_solver` keyword only?

FilippoAiraldi:
> Looks good :) Thank you!
> 
> Btw. is there a specific reason for making `n_batch_max`, and `num_threads_batch_solver` keyword only?

I usually require advance options/args as keywords. But I see that that's not the style you are using in `leap-c`, so I can remove it in some next PR

PR #119, opened 2025-09-05
set trainer to eval/train mode; wrap `validate` in `torch.inference_mode`

FilippoAiraldi:
In `Trainer.run`, PyTorch modules are never set to train or evaluation mode during training or validation, respectively. This can be an issue if layers such as DropOut are included in any submodules (see [this](https://docs.pytorch.org/docs/stable/notes/autograd.html#evaluation-mode-nn-module-eval)). This commit activates train mode (`self.train()`) or eval mode (`self.eval()`) as needed in `Trainer.run`.

On top of that, calls to `Trainer.validate` in `Trainer.run` have been wrapped in a `torch.inference_mode` context manager to disable gradient computations, hopefully speeding up validation runtime (see [this](https://docs.pytorch.org/docs/stable/notes/autograd.html#inference-mode)).

### DIFF
--- a/leap_c/trainer.py
+++ b/leap_c/trainer.py
@@ -204,17 +204,22 @@ class Trainer(ABC, nn.Module, Generic[TrainerConfigType]):
         train_loop_iter = self.train_loop()
 
         # initial policy validation
-        val_score = self.validate()
+        self.eval()
+        with torch.inference_mode():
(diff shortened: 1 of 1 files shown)


JasperHoffmann (approved):
Looks good!

PR #120, opened 2025-09-09
Missing dependency on PyYAML after removal of wand from dependencies

FilippoAiraldi:
Now that `wandb` is no longer a required dependency (which would have installed `PyYAML` automatically), we have to explicitly add `PyYAML` as a new dependency. It is used in `leap_c/trainer.py` to dump configurations as yaml files.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "leap-c"
 version = "2024.0.0"
-dependencies = ["gymnasium", "scipy", "numpy", "pandas"]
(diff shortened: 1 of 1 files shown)


dirkpr (approved):
Thanks! Ideally a test would  discover this missing dependency. We might want to discuss coverage.

FilippoAiraldi:
I agree. I discovered the bug via a fresh installation. 

I would expect any test that imports the trainer (or that leads to the trainer being invoked) to fail. But maybe PyYAML gets installed as a dependency of a test library, so the bug was missed?

PR #121, opened 2025-09-09
`episode_rollout` induces error with rendering

FilippoAiraldi:
In the current implementation, `episode_rollout` will always try to wrap the env in `RecordVideo` because the supplied `video_folder` is never `None`. This will induce an error if the environment does not support an image rendering mode.

Also, it will try to wrap the original env even if `render_episodes` is `0` or less, which should instead disable video recording as far as I can tell.

This PR fixes this issue by checking that the rendering mode is compatible with `RecordVideo`, i.e., `env.render_mode not in (None, "human", "ansi")` (I have taken these from `RecordVideo.__init__`).

The second commit improves type hinting and docstring of said method.

### DIFF
--- a/leap_c/utils/rollout.py
+++ b/leap_c/utils/rollout.py
@@ -38,15 +38,21 @@ def episode_rollout(
         keys "score", "length", "terminated", and "truncated" and a dictionary of
         policy statistics.
     """
-    if render_human and video_folder is not None:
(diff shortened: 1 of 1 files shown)


dirkpr (commented):
Thanks, looks good to me!

dirkpr (approved):

PR #127, opened 2025-09-11
Docs, typehints and refactorings

FilippoAiraldi:
This PR is a collection of many small changes to the following:

- docstrings: updated some docstrings to the new 100 char max line length. Also, formatted default values and new lines when inconsistent.
- type hints: added a bunch of type hints here and there, wherever I noticed them missing.
- minor refactorings.

No change to any functionality was made.

### DIFF
--- a/leap_c/ocp/acados/parameters.py
+++ b/leap_c/ocp/acados/parameters.py
@@ -1,7 +1,6 @@
-import warnings
-from collections.abc import Collection, Iterable
 from dataclasses import dataclass, field
-from typing import Any, Literal
+from typing import Any, Collection, Iterable, Literal
+from warnings import warn
 
 import casadi as ca
 import gymnasium as gym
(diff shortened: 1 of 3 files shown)


leap_c/torch/rl/sac_zop.py
```diff
@@ -350,10 +350,10 @@ def act(
 
     @property
     def optimizers(self) -> list[torch.optim.Optimizer]:
-        if self.alpha_optim is None:
-            return [self.q_optim, self.pi_optim]
-
-        return [self.q_optim, self.pi_optim, self.alpha_optim]
+        optimizers = [self.q_optim, self.pi_optim]
```
dirkpr, line 353: I sense some code duplication here. Should we refactor this?

dirkpr (approved):
Thanks a lot for modifying for type hints, python idioms, and cleaner code! Only left some minor comments for discussion.

PR #131, opened 2025-09-15
MX option in acados param manager

svsawant:
This pr contains a small change/functionality:
- Adding an option to AcadosParameterManager to generate casadi.MX variables (default is set to be SX as before)


dirkpr (changes requested):
Looks good! Just need to format the import block. I recommend using pre-commit, as described here: https://leap-c.github.io/leap-c/installation.html#linting-and-formatting

dirkpr (approved):
Looks good!

PR #132, opened 2025-09-15
Improving reproducibility

FilippoAiraldi:
This PR attempts to enforce reproducibility during validation and training. 

Essentially, the proposed approach consists in 1) creating a `np.random.Genrator` with the specified run seed and assigning it to the trainer, and 2) using said `np.random.Genrator` to generate a (hopefully, kinda unique) seed every time `env.reset` is called.

### DIFF
--- a/leap_c/utils/gym.py
+++ b/leap_c/utils/gym.py
@@ -40,7 +40,6 @@ def seed_env(
     Returns:
         tuple: The output of `env.reset`, i.e., the initial observation and info dictionary.
     """
-    out = env.reset(seed=seed, options=options)
     env.observation_space.seed(seed)
     env.action_space.seed(seed)
-    return out
+    return env.reset(seed=seed, options=options)
--- a/leap_c/utils/gym.py
+++ b/leap_c/utils/gym.py
@@ -1,13 +1,15 @@
-from typing import Callable, List
+from typing import Any, Callable, List, TypeAlias
 
-import gymnasium as gym
+from gymnasium import Env
 from gymnasium.core import ActType, ObsType
 from gymnasium.wrappers import OrderEnforcing, RecordEpisodeStatistics
 
-WrapperType = Callable[[gym.Env[ObsType, ActType]], gym.Env[ObsType, ActType]]
+WrapperType: TypeAlias = Callable[[Env[ObsType, ActType]], Env[ObsType, ActType]]
 
 
-def wrap_env(env: gym.Env, wrappers: List[WrapperType] | None = None) -> gym.Env:
+def wrap_env(
+    env: Env[ObsType, ActType], wrappers: List[WrapperType] | None = None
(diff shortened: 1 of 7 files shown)


leap_c/utils/gym.py
```diff
@@ -25,17 +27,20 @@ def wrap_env(env: gym.Env, wrappers: list[WrapperType] | None = None) -> gym.Env
...
         env: The environment to seed.
         seed: The seed to use.
+        options: Additional options to pass to `env.reset`.
 
     Returns:
-        gym.Env: The seeded environment.
+        tuple: The output of `env.reset`, i.e., the initial observation and info dictionary.
     """
-    env.reset(seed=seed)
+    out = env.reset(seed=seed, options=options)
```
dirkpr, line 43: Maybe more concise to implement as
```
env.observation_space.seed(seed)
env.action_space.seed(seed)
return env.reset(seed=seed, options=options)
```

leap_c/utils/gym.py
```diff
@@ -1,21 +1,23 @@
...
+) -> Env[ObsType, ActType]:
     """Wraps a gymnasium environment.
 
     Args:
         env: The environment to wrap.
         wrappers: A list of wrappers to apply to the environment.
 
     Returns:
-        gym.Env: The wrapped environment.
+        gymnasium.Env: The wrapped environment.
```
dirkpr, line 20: I think return type annotation and docstring are not consistent. Do you have a rule for when you import as `from gymnasium import Env` vs just using `gymnasium.Env`?

dirkpr (changes requested):
Looks god, only small comments!

leap_c/utils/gym.py
FilippoAiraldi: I do not discern; I am used to always annotate the full type.

In my projects, I usually include the return type for Sphinx to hyper-link it to the corresponding library, e.g., :method:`numpy.sum` or :class:`gymnasium.Env`. I think we should decide or whether to add args/return type hints, and if so, whether to create Sphinx hyper-links.

leap_c/utils/gym.py
FilippoAiraldi: No issue with that

leap_c/utils/gym.py
```diff
@@ -25,17 +27,19 @@ def wrap_env(env: gym.Env, wrappers: list[WrapperType] | None = None) -> gym.Env
...
 
     Returns:
-        gym.Env: The seeded environment.
+        tuple: The output of `env.reset`, i.e., the initial observation and info dictionary.
     """
-    env.reset(seed=seed)
     env.observation_space.seed(seed)
     env.action_space.seed(seed)
-    return env
+    return env.reset(seed=seed, options=options)
```
dirkpr, line 45: Nice!

PR #133, opened 2025-09-17
Fix bug from previous PR #127

FilippoAiraldi:
Previous PR #127 inadvertently introduced a bug by renaming `cfg.val_freq` to the older `cfg.val_interval` in `Trainer.run()`. 

This PR fixes the bug, and removes also another usage of the old `val_interval` that I found still lingering in the code base.

More noteworthy, no test picked this bug up. So no test is running `Trainer.run()` as of now, likely


dirkpr (approved):
Thank you for fixing!

PR #134, opened 2025-09-17
Improvements to logger

FilippoAiraldi:
In order of commits, this PR

- fixes the CSV file header; previously it was missing the first label associated to timestamps
- adds a test to make sure CSV logging is working as expected
- turns the logger into a context manager: it must be initialized via `with logger`. This ensures that resources are always closed properly (unless the program is killed externally)
- removes dependency on pandas to log data as CSV in favour of the built-in `csv` module.


JasperHoffmann (approved):
Looks good! Thank you :)

PR #141, opened 2025-09-23
Adapt HVAC example to new AcadosParameterManager and add preliminary rendering

dirkpr:
This PR adapts the hvac controller to the new `AcadosParameterManager`.
Also add rendering functionality for the `MatplotlibRenderEnv` using the `ctx` from the controller. Tested for validation rollouts.


JasperHoffmann (approved):
Looks good!

PR #142, opened 2025-09-24
Improve AcadosController

Confectio:
Previously, people could instantiate the AcadosController already (without it working), 
because it would just inherit the init of nn.Module. 
This PR changes the AcadosController such that it provides an init method, and changes the examples accordingly.
Also small fixes to typehints and documentation.


dirkpr (approved):
Looks good! Just a quick question in a comment.

PR #150, opened 2025-10-04
Make balance env more similar to the original balance env

Confectio:
Changed the `CarPoleBalanceEnv` to be a continuous version of gymnasiums Cart Pole env.


leap_c/examples/cartpole/env.py
```diff
@@ -402,12 +402,39 @@ def close(self):
 class CartPoleBalanceEnv(CartPoleEnv):
     """The same as the CartPoleEnv, but instead of swinging up, the pole starts in an upwards,
```
JasperHoffmann, line 403: Can we start with one sentence class summaries?

JasperHoffmann (approved):
Looks good! Class docstring could be slightly updated :)

PR #151, opened 2025-10-06
Simplify acados install

FreyJo:
such that options can be copied, and remove deprecated `-DACADOS_PYTHON=ON`


JasperHoffmann (approved):
Looks good!

PR #154, opened 2025-10-10
Feat: Add docker support

gursoyege:
Hi everyone, I’ve created a Docker container for leap-c that can be built directly using VS Code Dev Containers with one click. The setup is described in the README. It supports Windows, macOS, and Debian (though I only tested on Ubuntu), with both CPU and GPU options for PyTorch.

Detailed instructions are in the readme but basically to use it:

Clone the leap-c repo into an empty folder in host (e.g., syscop_ws).

Open "leap-c/docker" folder with  “Open in Container” in vscode (bottom left corner >< button)

Choose your install option (force CPU, force GPU, or Auto).

The container mounts syscop_ws, so changes are synced between host and container.

Feel free to use or share it. It should save time for anyone setting up the environment.

### DIFF
--- a/docker/.devcontainer/.gitignore
+++ b/docker/.devcontainer/.gitignore
@@ -1,0 +1,4 @@
+# Auto-generated devcontainer configuration
+devcontainer.local.json
+
+
--- a/docker/.devcontainer/select-target.sh
+++ b/docker/.devcontainer/select-target.sh
@@ -1,0 +1,126 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Detect OS
+detect_os() {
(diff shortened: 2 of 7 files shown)


JasperHoffmann:
@gursoyege Sorry for the delayed response, I tested it once and it seemed to work. Will test it next week again, trying to get his merged :)

FranekStark:
Hi @JasperHoffmann , was is the status of that? - I think it would be very helpfull :)

JasperHoffmann:
> Hi @JasperHoffmann , was is the status of that? - I think it would be very helpfull :)

Yes then I will add it next Monday :P Have some old code lying around but need to check it again.

JasperHoffmann:
> Hi @JasperHoffmann , was is the status of that? - I think it would be very helpfull :)

We now just implemented it and pre-built images are now available ;) You can find it here #320

PR #161, opened 2025-10-27
Int bug in MLP bug and small refactoring

FilippoAiraldi:
Currently, in `leap_c.torch.nn.Mlp.__init__`, if `input_sizes` or `output_sizes` is of integer type but not explicitly an `int` (e.g., `np.int32`, `np.int64`, etc.), the argument is not correctly handled, i.e., it is regarded as a `list[int]` instead.

In order of commits, this PR

1. switches from `isinstance(..., int)` to `isinstance(..., Integral)` to increase support for a larger spectrum of integer types
2. adds missing docstrings and adjust the old ones to the new 100 char line max width
3. removes unused `Mlp` attributes, and simplifies the creation of the MLP layers via `itertools.pairwise`
4. adds an additional test for `Mlp`.

### DIFF
--- a/leap_c/torch/nn/mlp.py
+++ b/leap_c/torch/nn/mlp.py
@@ -2,7 +2,7 @@ from collections.abc import Sequence
 from dataclasses import dataclass
 from itertools import pairwise
 from numbers import Integral
-from typing import Callable, Iterable, Literal
+from typing import Callable, Iterable, Literal, get_args
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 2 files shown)


leap_c/torch/nn/mlp.py
```diff
@@ -18,34 +31,50 @@ def string_to_activation(activation: Activation) -> nn.Module:
...
+
+    Returns:
+        Callable[[nn.Module], None]: The weight initialization function that takes in a module and
+            initializes its weights, returning nothing.
+    """
     if weight_init == "orthogonal":
         return orthogonal_init
-    else:
-        raise ValueError(f"Weight initialization {weight_init} not recognized.")
+    raise ValueError(f"Weight initialization {weight_init} not recognized.")
```
dirkpr: Return the list of available weight_init?

FilippoAiraldi: Done

JasperHoffmann (approved):

PR #162, opened 2025-10-29
HVAC modifications

dirkpr:
The hvac example needs modifications to address the following shortcomings:
1) The heat power can be negative, allowing the agent to cool the room and sell "extracted" energy back to the grid to get a reward. This behavior is not realistic.
2) Rendering was not finished previously
3) The reset method allowed states outside of the heating period when the trivial action would be to not heat.

This PR introduces these modifications.

### DIFF
--- a/leap_c/examples/hvac/env.py
+++ b/leap_c/examples/hvac/env.py
@@ -568,7 +568,7 @@ class StochasticThreeStateRcEnv(MatplotlibRenderEnv):
             counter = 0
             date_valid = False
             while not date_valid and counter < n_max:
-                idx = self._np_random.integers(low=min_start_idx, high=max_start_idx + 1)
+                idx = self.np_random.integers(low=min_start_idx, high=max_start_idx + 1)
                 date = self.data.index[idx]
                 if date.month in [1, 2, 3, 4, 9, 10, 11, 12]:
                     date_valid = True


leap_c/examples/hvac/env.py
```diff
@@ -561,7 +561,26 @@ def reset(
...
             max_start_idx = len(self.data) - self.N_forecast - self.max_steps + 1
-            self.idx = self.np_random.integers(low=min_start_idx, high=max_start_idx + 1)
+
+            self.idx = min_start_idx
+
+            n_max = 1000
+            counter = 0
+            date_valid = False
+            while not date_valid and counter < n_max:
+                idx = self.np_random.integers(low=min_start_idx, high=max_start_idx + 1)
```
Confectio: Do you know where this object `self.np_random` is created and seeded? I only know that `gym.Env.reset(seed=seed)` seeds `self._np_random`.

leap_c/examples/hvac/env.py
```diff
@@ -561,7 +561,26 @@ def reset(
...
+                date = self.data.index[idx]
+                if date.month in [1, 2, 3, 10, 11, 12]:
```
Confectio: Is the heating period really so restrictive? E.g., I would also count the April in Germany, and depending on the data, e.g., countries more to the north, it might include even more months?

Confectio (changes requested):
Maybe make Summer feasible by making cost dependend on abs(action)?

Confectio (approved):
Looks good 👍

PR #168, opened 2025-11-06
BUG FIX: test skipped because controller names are now in PLANNER_REGISTRY

FilippoAiraldi:
Tests were skipped because CONTROLLER_REGISTRY is now empty. This begs the question whether the word `controller` should be replaced by `planner` in the whole test file.

### DIFF
--- a/tests/leap_c/test_scripts.py
+++ b/tests/leap_c/test_scripts.py
@@ -6,10 +6,11 @@ from pathlib import Path
 import pytest
 
 import leap_c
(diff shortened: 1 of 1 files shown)


Confectio (approved):
Again, nice catch👍 . I was wondering just yesterday, why the code-coverage suddenly dropped by 5%, but then didn't give it enough thought :'D.

I would not rename it for now, because all the current scripts actually only need a controller, and even though we do not have any yet, there could be controllers implemented that are not planners (planners are a subset of controllers).

PR #170, opened 2025-11-06
`ctx` interface and type hinting

FilippoAiraldi:
This PR introduces a minimal interface to what constitutes a context, allowing for minimal requirements and some level of type hinting of said objects.

The core of the PR is the introduction of a `Protocol` (i.e., a Python's interface) that defines what a context should look like. The various classes that handle these contexts, e.g., `ParameterizedController` and the newer `ParameterizedPlanner`, are made generic w.r.t. said protocol. This approach allows to

1. easily modify what `ctx` should look like by adding or removing fields, methods or properties from `CtxInterface`
2. new implementations have a template to refer to that specifies what needs to satisfied (alongside the concrete implementation `AcadosDiffMpcCtx`, which remain still valid of course)
3. finally, we can add more specific type hints for `ctx` objects other than `typing.Any`

The first two commits in the PR are minor; I separated them from the rest mainly to reduce clutter.

1. adds missing type hints and formats docstrings in all the files involved in the PR
2. replaces the extensive if/else branches in the new `AcadosPlanner` with a dictionary-based solution, and adds a test to make sure no option is missed 

The other commits read as follow:

3. implements the interface `CtxInterface`, which is then used in `ParameterizedController` and `ParameterizedPlanner`. `AcadosPlanner` further specializes its interface to be a subtype of `AcadosDiffMpcCtx`.
4. extends the new `CtxType` (which is bound to `CtxInterface`) to the SAC-MPC agents and relevant classes
5. finally, extends the new `CtxType` to the examples.

### DIFF
--- a/leap_c/controller.py
+++ b/leap_c/controller.py
@@ -2,7 +2,7 @@
 PyTorch."""
 
 from abc import ABCMeta, abstractmethod
-from typing import Callable, Generic, Protocol, TypeVar, Union
+from typing import Any, Callable, Generic, Protocol, TypeVar, Union
 
 import gymnasium as gym
 from numpy import ndarray
@@ -43,14 +43,12 @@ class ParameterizedController(nn.Module, Generic[CtxType], metaclass=ABCMeta):
     collate_fn_map: dict[Union[type, tuple[type, ...]], Callable] | None = None
 
     @abstractmethod
-    def forward(
-        self, obs: Tensor, param: Tensor, ctx: CtxType | None = None
-    ) -> tuple[CtxType, Tensor]:
+    def forward(self, obs: Any, param: Any, ctx: CtxType | None = None) -> tuple[CtxType, Tensor]:
         """Computes action from observation, parameters and internal context.
 
         Args:
-            obs (Tensor): Observation input to the controller (e.g., state vector).
-            param (Tensor): Parameters that define the behavior of the controller.
+            obs: Observation input to the controller (e.g., state vector).
+            param: Parameters that define the behavior of the controller.
             ctx (CtxType, optional): Optional internal context passed between invocations.
 
         Returns:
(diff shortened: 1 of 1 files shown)


JasperHoffmann:
I really love protocols, and originally we planned to provide it, like you did, but in this case I am not sure, how much the user will get out of it :) Maybe we can discuss in the next sync!

FilippoAiraldi:
To be fair, from the user-code point of view, type hints are not a necessity, i.e., if a user wants not to do it, `leap-c` would still works fine. So I agree with you there.

From a developer point of view, however, it is nice to have the library's interface properly type-hinted, so that its I/O behaviour are explicitly stated.

As usual, this PR is just a suggestion; if we deem it not necessary, we can skip it :)

leap_c/controller.py
```diff
@@ -1,83 +1,104 @@
...
+            custom data types. For more information, please refer to, e.g.,
+            https://docs.pytorch.org/docs/stable/data.html#torch.utils.data.default_collate.
     """
 
     collate_fn_map: dict[Union[type, tuple[type, ...]], Callable] | None = None
 
     @abstractmethod
-    def forward(self, obs, param, ctx=None) -> tuple[Any, torch.Tensor]:
+    def forward(
+        self, obs: Tensor, param: Tensor, ctx: CtxType | None = None
```
JasperHoffmann: I think obs and param do not necessarily need to be a Tensor :) Especially obs! So I would suggest remove the typehint here!

JasperHoffmann (changes requested):
Looks good! I would remove the type hint for obs. Obs will be pretty soon not a tensor anymore.

leap_c/planner.py
```diff
@@ -1,115 +1,116 @@
...
+            u0 (Tensor, optional): The computed initial control action. Is `None` if
                 the action is already provided.
-            x: The computed sequence of states (torch.Tensor).
-                Expected shape (N+1, *state_dims).
-            u: The computed sequence of controls (torch.Tensor).
-                Expected shape (N, *control_dims).
-            value: The cost value of the computed trajectory (torch.Tensor).
+            x (Tensor): The computed sequence of states. Expected shape `(N+1, *state_dims)`.
+            u (Tensor): The computed sequence of controls. Expected shape `(N, *control_dims)`.
+            value (Tensor): The cost value of the computed trajectory.
```
FilippoAiraldi: The docstring suggests there should be a fifth returned object; however, the type hints only list four. Which one should was the intended one?

PR #171, opened 2025-11-07
Bugfix: Update part of the train loop was not covered by ``test_scripts.py``

Confectio:
Follow-up to PR #169 , where it was noticed that we probably do not test the update part of the train_loops somehow.
This has been confirmed by running ``test_scripts.py`` manually with the debugger.
The PR overrides the config for the tests such that the update part of the train loop runs.

Also fixes a local issue with my pytest test recovery for vscode specifically in ``test_scripts.py`` .

### DIFF
--- a/tests/leap_c/test_scripts.py
+++ b/tests/leap_c/test_scripts.py
@@ -137,6 +137,9 @@ def _run_script(script, tmp_path, env, reuse_code_dir, controller=None):
     cfg = create_cfg(script, env=env, controller=controller)
     cfg.log = False
     cfg.trainer.train_steps = 10
+    cfg.trainer.train_start = 0
+    cfg.trainer.update_freq = 1
+    cfg.trainer.batch_size = 8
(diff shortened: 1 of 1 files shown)


tests/leap_c/test_scripts.py
```diff
@@ -8,7 +8,10 @@
 import leap_c
 from leap_c.examples import CONTROLLER_REGISTRY, ENV_REGISTRY, PLANNER_REGISTRY
 
-LEAP_C_ROOT = Path(leap_c.__file__).resolve().parent.parent
+if getattr(leap_c, "__file__", None):
+    LEAP_C_ROOT = Path(leap_c.__file__).resolve().parent.parent
+else:
+    LEAP_C_ROOT = Path(next(iter(leap_c.__path__))).resolve().parent
```
JasperHoffmann: Why do we need this if/else distinction?

PR #175, opened 2025-11-12
Include pydocstyle rules for ruff lint.

dirkpr:
Linting rules for pydocstyle are not enforced.

This PR includes the ruleset D in the linting options (https://docs.astral.sh/ruff/rules/#pydocstyle-d) and works on enforcing them through ruff when submitting a PR. The rules D100-107 are ignored (undocumented classes, methods, etc...)

### DIFF
--- a/leap_c/torch/nn/bounded_distributions.py
+++ b/leap_c/torch/nn/bounded_distributions.py
@@ -28,7 +28,7 @@ class BoundedDistribution(nn.Module):
     def forward(
         self, *defining_parameters, deterministic: bool = False
     ) -> tuple[torch.Tensor, torch.Tensor, dict[str, float]]:
-        """Samples from the distribution.
+        """Sample from the distribution.
 
         If `deterministic` is True, the mode of the distribution is used instead of
         sampling.
@@ -163,7 +163,7 @@ class SquashedGaussian(BoundedDistribution):
     def forward(
         self, mean: torch.Tensor, log_std: torch.Tensor, deterministic: bool = False
     ) -> tuple[torch.Tensor, torch.Tensor, dict[str, float]]:
-        """Forward pass.
+        """Sample from the SquashedGaussian distribution.
 
         Args:
             mean: The mean of the normal distribution.
--- a/leap_c/torch/rl/sac.py
+++ b/leap_c/torch/rl/sac.py
@@ -164,7 +164,7 @@ class SacActor(nn.Module):
     def forward(
         self, obs: torch.Tensor, deterministic: bool = False
     ) -> tuple[torch.Tensor, torch.Tensor, dict[str, float]]:
-        """Actor forward pass.
+        """Sample actions from the policy given observations.
 
         The given observations are passed to the extractor to obtain features.
         These are used by the MLP to predict parameters used to define a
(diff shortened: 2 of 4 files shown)


leap_c/examples/hvac/env.py
```diff
@@ -57,14 +57,18 @@ def __init__(
...
+        """Initialize the stochastic environment.
 
         Args:
             params: Dictionary of thermal parameters
             step_size: Time step for the simulation in seconds
-            ambient_temperature_function: Function for ambient temperature
-            solar_radiation_function: Function for solar radiation
+            start_time: Start time for the historical data
+            horizon_hours: Prediction horizon in hours
+            max_hours: Maximum simulation time in hours
```
Confectio: Until an episode ends?

leap_c/examples/hvac/env.py
```diff
@@ -57,14 +57,18 @@ def __init__(
...
+            render_mode: Render mode for the environment
+            price_zone: Price zone for electricity prices
```
Confectio: What are the possible values? Maybe create a type for it (Literal[....])?

leap_c/torch/nn/bounded_distributions.py
```diff
@@ -160,7 +163,8 @@ def __init__(
     def forward(
         self, mean: torch.Tensor, log_std: torch.Tensor, deterministic: bool = False
     ) -> tuple[torch.Tensor, torch.Tensor, dict[str, float]]:
-        """
+        """Forward pass.
```
Confectio: "Sample from the SquashedGaussian distribution"?

leap_c/torch/rl/sac.py
```diff
@@ -163,7 +164,9 @@ def __init__(
     def forward(
         self, obs: torch.Tensor, deterministic: bool = False
     ) -> tuple[torch.Tensor, torch.Tensor, dict[str, float]]:
-        """The given observations are passed to the extractor to obtain features.
+        """Actor forward pass.
```
Confectio: Be more expressive, e.g., "Calculates the action to take in the given observation and its log_prob" or sth like this?

leap_c/torch/rl/sac_fop.py
```diff
@@ -243,7 +246,9 @@ def __init__(
     def forward(
         self, obs: np.ndarray, ctx: CtxType | None = None, deterministic: bool = False
     ) -> SacFopActorOutput:
-        """The given observations are passed to the extractor to obtain features.
+        """Forward pass.
```
Confectio: Same as for SacActor?

Confectio:
Nice 👍 I have a few comments though :)

leap_c/torch/rl/sac.py
Confectio: Yeah I know, but in my opinion it should contain more than what the name of the function already says

Confectio (approved):
Just for the future: I think the addition of the return values in the docstring is overkill, when we have a dataclass dedicated to explaining it already and we do not add more information to that. E.g., the output of the SacFopActor forward returns SacFopActorOutput. But of course since we already have it now we can leave it like this :) .

PR #176, opened 2025-11-13
Hvac cleanup

dirkpr:
The HVAC example includes a lot of unused code and its structure is outdated since 8360cd8456c1990e69cb83915094b8e3742a40a1.

This PR:
- removes unused code and updates the structure to use the `AcadosPlanner` interface and match the other examples.
- refactors the hvac example to use `class Dataset` to interact with historical environmental data. Define the dynamics in `dynamics.py` and forecasts in `class Forecaster`.
- extends the flexibility of `HvacPlanner` to declare subsets of the parameters as learnable/non-learnable
- Modifies the github CI pipeline to use smaller jobs and split into integration tests and units tests.

### DIFF
--- a/leap_c/ocp/acados/parameters.py
+++ b/leap_c/ocp/acados/parameters.py
@@ -5,6 +5,7 @@ from warnings import warn
 import casadi as ca
 import gymnasium as gym
 import numpy as np
+import torch
 from acados_template import AcadosOcp
 from casadi.tools import entry, struct, struct_symMX, struct_symSX
 
@@ -436,6 +437,30 @@ class AcadosParameterManager:
 
         learnable_param_names = self.learnable_parameters.keys()
         return any(fnmatch.fnmatch(name, pattern) for name in learnable_param_names)
+
+    def get_structured_learnable_parameters(
+        self,
+        param_values: np.ndarray | torch.Tensor,
+        label=None,
+    ) -> struct:
+        """Get a structured representation of the learnable parameters from flat values.
+
+        Args:
+            param_values: Flat numpy array of learnable parameter values.
+            label: Optional substring to filter parameters by name.
+
+        Returns:
+            A CasADi struct with the learnable parameters populated from param_values.
+        """
+        # Get indices of learnable parameters that include the substring 'label' if provided
+        if label is not None:
+            matching_keys = [key for key in self.learnable_parameters.keys() if label in key]
+        else:
+            matching_keys = list(self.learnable_parameters.keys())
+
+        print(matching_keys)
+
+        return None
 
 
 def _define_starts_and_ends(end_stages: list[int], N_horizon: int) -> tuple[list[int], list[int]]:


leap_c/examples/hvac/env.py
```diff
@@ -26,13 +26,6 @@
...
-# Constants
-DAYLIGHT_START_HOUR = 6
-DAYLIGHT_END_HOUR = 18
-MEAN_AMBIENT_TEMPERATURE = convert_temperature(0, "celsius", "kelvin")
-MAGNITUDE_AMBIENT_TEMPERATURE = 5
-MAGNITUDE_SOLAR_RADIATION = 200
-
 
 class StochasticThreeStateRcEnv(MatplotlibRenderEnv):
     """Simulator for a three-state RC thermal model with exact discretization of Gaussian noise.
```
Confectio: This is by far the most complex environment, but it has the sparsest description (see other envs). This docstring should be comparable to the other envs. 
On top, we could also add a reference to the name of a paper in the docstring, if it makes sense and an appropriate paper exists.

Confectio: Possibly missing the Attribute descriptions and/or typehints, depending on the outcome of our currently ongoing discussion

leap_c/examples/hvac/planner.py
```diff
@@ -0,0 +1,332 @@
...
+        stagewise: Whether to use stage-wise parameters for forecasts
+            (ambient temperature, solar radiation, prices).
+    """
+
+    N_horizon: int = 96  # 24 hours in 15 minutes time steps
+    stagewise: bool = False
+
+
+class HvacPlanner(AcadosPlanner[HvacPlannerCtx]):
+    """acados-based planner for the HVAC system.
```
Confectio: I like this docstring :)

leap_c/examples/hvac/planner.py
```diff
@@ -0,0 +1,332 @@
...
+        if isinstance(Ta_forecast, torch.Tensor):
+            Ta_forecast = Ta_forecast.cpu().numpy().flatten()
+        if isinstance(solar_forecast, torch.Tensor):
+            solar_forecast = solar_forecast.cpu().numpy().flatten()
+        if isinstance(price_forecast, torch.Tensor):
+            price_forecast = price_forecast.cpu().numpy().flatten()
+
+        N_horizon = self.diff_mpc.ocp.solver_options.N_horizon
+        for stage in range(N_horizon + 1):
+            param[f"Ta_{stage}_{stage}"] = Ta_forecast[stage]
```
Confectio: What happens if this key does not exist in param? Will it throw an error (good)? 
Or will it enlarge the parameter structure (bug)?

I could be wrong, but these 3 lines seem eronneous to me.
In particular they possibly assume:
1. If Ta, Phi or price is learnable, then all 3 are learnable.
2. The naming of stagewise learnable parameters inside the parameter manager stays the same for all time.
3. They are always stagewise per stage and not over a span of stages.

I think we should get rid of these assumptions, 
or if not possible check these conditions and throw an error they are not met.

Confectio (commented):
Looks good, but I have a few comments :)

leap_c/examples/hvac/acados_ocp.py
```diff
@@ -0,0 +1,276 @@
...
+    Args:
+        param_manager: The parameter manager containing the parameters for the OCP.
+        N_horizon: Number of time steps in the horizon.
+        name: Name of the OCP model.
+        x0: Initial state. If None, a default value is used.
+
+    Returns:
+        AcadosOcp: The configured OCP object.
+    """
+    dt: float = 900.0  # Time step in seconds (15 minutes)
```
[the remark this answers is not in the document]
dirkpr: From all practical aspects, it makes sense to keep it fixed at 15 min. The price resolution is 15 min, and the related sensors also often aggregate in a way that 15 min is a good sampling time.

leap_c/examples/hvac/planner.py
```diff
@@ -0,0 +1,332 @@
...
+            ctx=diff_mpc_ctx,
+        )
+
+        ctx = HvacPlannerCtx(
+            diff_mpc_ctx,
+            qh=x[:, 1, 3].detach(),
+            dqh=x[:, 1, 4].detach(),
+        )
+
+        return ctx, u0, x, u, value
```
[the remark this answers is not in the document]
dirkpr: Good catch! I am surprised it was working so far. Will change it to `qh`.

leap_c/examples/hvac/planner.py
dirkpr: This point is somewhat related to your previous remark https://github.com/leap-c/leap-c/pull/176#discussion_r2523957547 where I argue for keeping the time interval fixed at 15min intervals (900s). We can assume to have forecasts for all these three quantities. And I think for this example it is okay to have all 3 learnable or all three not learnable.

leap_c/examples/hvac/planner.py
```diff
@@ -0,0 +1,332 @@
...
+        - quarter_hour: Current quarter hour of the day (0-95)
+        - day_of_year: Current day of the year (1-365)
+        - Ti: Indoor air temperature in Kelvin
+        - Th: Radiator temperature in Kelvin
+        - Te: Envelope temperature in Kelvin
+        - Ta_forecast: Ambient temperature forecast for the next N steps
+        - solar_forecast: Solar radiation forecast for the next N steps
+        - price_forecast: Electricity price forecast for the next N steps
+    """
+    if obs.ndim > 1:
```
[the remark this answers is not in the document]
dirkpr: Removed the whole thing.

leap_c/examples/hvac/planner.py
Confectio: I mean it can make sense to only learn individual forecasts, but I agree that this change can be made once someone actually needs it. 

So what I wanted to say was not that we should definitely not assume them, but to make it noticeable, when this assumption is not given anymore, and the system therefore behaves in a way the user might not anticipate. 

E.g., If I just quickly change the default hvac params, which lie in another file even, because I want to make the learning problem easier and therefore only forecast prices first, then I will probably get behaviour that is not anticipated, because at some point far away (here) it was assumed that I would not do this. It would be good to therefore throw an expressive exception here and add it to the docstring, to tell me I should not do this when I run my code. 

The same holds for the other two assumptions I listed.

leap_c/examples/hvac/env.py
```diff
@@ -32,6 +32,58 @@ class StochasticThreeStateRcEnv(MatplotlibRenderEnv):
...
+        +1 if lb <= Ti <= ub
+        0  else
+
+    The normalized energy costs reward is computed as
+        r_costs = -(action/action_max) * (price/price_max)
+
+
+    Terminates:
+    -----------
+    Terminates after max_hours is reached in simulated time or when the historical data ends.
```
Confectio: As far as I can see, the env truncates in that case, but never terminates

leap_c/examples/hvac/planner.py
```diff
@@ -221,7 +221,7 @@ def forward(
             dqh=x[:, 1, 4].detach(),
         )
 
-        return ctx, u0, x, u, value
+        return ctx, qh, x, u, value
```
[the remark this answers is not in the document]
Confectio: I am confused. I can only see the lines
``qh = torch.zeros((batch_size, 1), dtype=torch.float64, device=obs.device)``
and
``qh = ctx.qh``
at the beginning of the function, therefore it should still have this value?
Aside from this, I thought the point is that qh is fixed for the initial state (because our diff function automatically fixes the whole initial state and qh is part of the state here), therefore we output the qh of the next step (after applying u0), which would be x[:, 1, 3].detach().
But now that I think about it, this would also be a bug I think. Since qh only evolves with dqh for one timestep, the optimal action chosen for ddqh would not reach qh after one step. Therefore we would have to take x[:, 2, 3].detach() or something?

leap_c/examples/hvac/env.py
dirkpr: Fixed

leap_c/examples/hvac/env.py
```diff
@@ -32,6 +32,58 @@ class StochasticThreeStateRcEnv(MatplotlibRenderEnv):
...
+
+    Truncates:
+    ----------
+    Same as termination.
+
```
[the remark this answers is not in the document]
dirkpr: Comes at a later stage

JasperHoffmann (commented):
Just added two high level comments :)

leap_c/examples/hvac/planner.py
```diff
@@ -239,12 +239,41 @@ def default_param(self, obs: np.ndarray | None) -> np.ndarray:
...
+        else:
+            # Batched observation case: obs has shape (n_batch, obs_dim)
+            n_batch = obs.shape[0]
+            forecasts = obs[:, 5:].reshape(n_batch, 3, -1).transpose(0, 2, 1)
+
+            # Get n_param from the flattened default parameters
+            n_param = len(param.cat.full().flatten())
+            batch_param = np.zeros((n_batch, n_param))
+
+            for i in range(n_batch):
```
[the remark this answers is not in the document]
dirkpr: Agreed. We will need to rewrite the param manager to be more efficient here.

leap_c/examples/hvac/config.py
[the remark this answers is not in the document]
FilippoAiraldi: if the circular import is caused by importing for type annotation, you can always circumvent the issue by guarding the import with TYPE_CHECKING

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    import config 
```

leap_c/examples/hvac/planner.py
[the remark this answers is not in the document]
dirkpr: Just to be clear, `x[:, 1, 3].detach()` is what the planner sees will be the optimal `qh` in 15 min, applied at the controller update (0 min).

leap_c/examples/hvac/acados_ocp.py
```diff
@@ -6,8 +6,10 @@
 from acados_template import ACADOS_INFTY, AcadosOcp
 from scipy.constants import convert_temperature
 
-from leap_c.examples.hvac.config import BestestHydronicParameters
-from leap_c.examples.hvac.util import transcribe_discrete_state_space
+from leap_c.examples.hvac.dynamics import (
```
dirkpr: What is the reason for the name change? The param name `BestestHydronicParameters` were to indicate their connection to this test case: https://ibpsa.github.io/project1-boptest/testcases/ibpsa/testcases_ibpsa_bestest_hydronic/

leap_c/examples/hvac/planner.py
dirkpr: Yes, the are always stagewise per stage, I added a note in the docstring and expection when this is not these case. As for varying learnable non-learnable that is also added. The naming will stay the same for all time.

leap_c/examples/hvac/acados_ocp.py
JasperHoffmann: Do you still want to rename it? From my side the docstring is enough :)

.github/workflows/tests.yml
```diff
@@ -8,83 +8,80 @@ on:
...
 
-  run_tests:
+  script_tests:
+    name: Integration Tests (${{ matrix.test-group.env }})
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.11", "3.12", "3.13"]
+        python-version: ["3.12"]
```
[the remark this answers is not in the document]
dirkpr: Changed in https://github.com/leap-c/leap-c/pull/176/commits/d77dae9f75de7da70255627588047c384d26adce

JasperHoffmann (approved):
Looks good!

PR #186, opened 2025-11-27
Include all dataset bounds in min max and use them in rendering and obs bounds

dirkpr:
This PR modifies the hvac dataset and environment to pick the bounds for observation space and rendering axes from the provided dataset.

### DIFF
--- a/leap_c/examples/hvac/dataset.py
+++ b/leap_c/examples/hvac/dataset.py
@@ -70,7 +70,7 @@ class HvacDataset:
                 If data is None, cfg is used to load data from files.
 
         Note:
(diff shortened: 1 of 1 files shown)


leap_c/examples/hvac/dataset.py
```diff
@@ -59,7 +59,6 @@ class HvacDataset:
     def __init__(
         self,
         data: pd.DataFrame | None = None,
-        price_max: float | None = None,
```
Confectio: The docstring needs an update

dirkpr: Good catch. Done!

PR #193, opened 2025-12-02
Mode concentration beta

dirkpr:
This PR introduces a new `ModeConcentrationBeta` distribution and improves the LQR example to a mass-spring damper

### New Beta Distribution

Adds `ModeConcentrationBeta` to the available `BoundedDistribution` options. This distribution:

- Accepts `mode` and `log_concentration` in the forward pass (analogous to `mean` and `log_std` in `SquashedGaussian`)
- Respects parameter/action bounds through a soft sigmoid before sampling
- Avoids potential issues with probability mass accumulating at the bounds

**Usage:** Set `cfg.trainer.actor.distribution_name = "mode_concentration_beta"`

### LQR Example Improvements

- Added slack variables for state constraints
- Implemented Cholesky factorization of the cost matrix
- Extended the dynamic model to include mass, damping, and stiffness parameters
- Introduced parameter mismatch between the environment and planner models

### DIFF
--- a/leap_c/torch/nn/bounded_distributions.py
+++ b/leap_c/torch/nn/bounded_distributions.py
@@ -390,15 +390,6 @@ class ModeConcentrationBeta(BoundedDistribution):
             An output sampled from the ModeConcentrationBeta, the log probability of this output
             and a statistics dict.
         """
-        # Mode must be in [padding, 1-padding]
-        mode = self.padding + (1.0 - 2 * self.padding) * torch.sigmoid(mode)
-
-        # Concentration must be > 2 for valid parameterization
-        log_conc = self.log_conc_min + (self.log_conc_max - self.log_conc_min) * torch.sigmoid(
-            log_conc
-        )
-        concentration = torch.exp(log_conc)
-
         if anchor is not None:
             # Convert anchor to tensor if it's a numpy array
             if not isinstance(anchor, torch.Tensor):
@@ -408,6 +399,15 @@ class ModeConcentrationBeta(BoundedDistribution):
 
             inv_anchor = self.inverse(anchor)
             mode = mode + inv_anchor  # Use out-of-place operation to avoid modifying view
+
+        # Mode must be in [padding, 1-padding]
+        mode = self.padding + (1.0 - 2 * self.padding) * torch.sigmoid(mode)
+
+        # Concentration must be > 2 for valid parameterization
+        log_conc = self.log_conc_min + (self.log_conc_max - self.log_conc_min) * torch.sigmoid(
+            log_conc
+        )
+        concentration = torch.exp(log_conc)
 
         if deterministic:
             # Deterministic: just use the mode scaled to action space


leap_c/examples/lqr/planner.py
```diff
@@ -21,15 +21,9 @@ class LqrPlannerConfig:
         N_horizon: The number of steps in the MPC horizon.
             The MPC will have N+1 nodes (the nodes 0...N-1 and the terminal
             node N).
-        T_horizon: The simulation time between two MPC nodes will equal
-            T_horizon/N_horizon [s] simulation time (currently uses N_horizon
-            as the time horizon in acados_ocp.py).
-        dtype: Type the planner output tensors will automatically be cast to.
```
Confectio: Dtype should be in the dostring

leap_c/torch/nn/bounded_distributions.py
```diff
@@ -313,3 +315,190 @@ def forward(
...
+        mode: torch.Tensor,
+        log_conc: torch.Tensor | None = None,
+        deterministic: bool = False,
+        anchor: torch.Tensor | None = None,
+    ) -> tuple[torch.Tensor, torch.Tensor, dict[str, float]]:
+        """Sample from the ModeConcentrationBeta distribution.
+
+        Args:
+            mode: The mode of the Beta distribution in [0, 1] space.
+            log_conc: The logarithm of the concentration parameter,
```
Confectio: Why is there an option to make log_conc equal None? I would prefer to just always pass log_conc and if I want determinism, pass deterministic=True, like in squashed gaussian?

leap_c/torch/nn/bounded_distributions.py
```diff
@@ -313,3 +315,190 @@ def forward(
...
+        concentration = torch.as_tensor(concentration)
+
+        # Compute alpha, beta from mode and total concentration c
+        alpha = 1.0 + mode * (concentration - 2.0)
+        beta = 1.0 + (1.0 - mode) * (concentration - 2.0)
+
+        # Update the internal Beta distribution with new parameters
+        self._beta_dist = Beta(concentration1=alpha, concentration0=beta, validate_args=None)
+
+    def _log_prob(self, value: torch.Tensor) -> torch.Tensor:
```
Confectio: Seems unnecessary to me (its not used?) => remove this function?

leap_c/torch/nn/mlp.py
```diff
@@ -156,3 +156,41 @@ def forward(self, *x: torch.Tensor) -> torch.Tensor | tuple[torch.Tensor, ...]:
...
+
+        Returns:
+            mode: Mode parameter in [0, 1], shape (..., output_dim)
+            concentration: Concentration parameter > 2, shape (..., output_dim)
+        """
+        # Get raw outputs from parent MLP
+        mode_raw, concentration_raw = super().forward(x)
+
+        # Mode must be in [0, 1]
+        mode = torch.sigmoid(input=mode_raw)
```
Confectio: I would prefer to delete this class and to enforce the bounds in the Beta distribution itself (compare to ScaledBeta, where we just add +1 automatically, after exponentiating the logs). As an alternative, we should at least just check the bounds in the Beta distribution and throw an error if the input is illegal.

Confectio (changes requested):
Nice, I like the possibility to use mode and concentration of a beta distribution very much :).
There are a few small remarks though.

scripts/run_sac_zop_beta.py
```diff
@@ -0,0 +1,158 @@
+"""Main script to run SAC-ZOP experiments."""
```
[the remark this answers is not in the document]
dirkpr: Was a leftover, removed it.

PR #198, opened 2025-12-03
Fixed device bug in hvac planner

JasperHoffmann:
Fixed device bugs on hvac.


Confectio (commented):
Nice, but one comment :)

PR #207, opened 2025-12-10
Hvac closed loop rendering

dirkpr:
This PR modifies the rendering for the `hvac` example such that the the exploration is visualized for the current param interface with learnable `ref_Ti` and `q_Ti`

output


JasperHoffmann (approved):
Looks good, could we maybe additionally add a date and time in the right column?

PR #212, opened 2026-01-09
Improve gradient testing

Confectio:
Right now the gradient functionality is only tested on LMPC examples. 
Motivated by #211 , this PR adds more gradient tests with indefinite full hessian (cartpole example) which get triggered by this bug (and hopefully other bugs that might only show on NMPC).

Also:
- Turned on stagewise parameter testing (was turned off previously because one entry was off). This was caused by finite differences going over a large jump in the Value function, leading to one very different gradient entry than the analytical gradient => Fixed by changing x0 (and hence, the test samples) to lie appropriately between ubx and lbx.
- Turned off default casting of output tensors to float32 in the autograd function. Instead no cast is done there, the cast has been moved to the diff_mpc. This lets people use the float64 accuracy of the MPC output if needed, through an input in the constructor/a flag in the examples.
- Improved accuracy and tolerance in the gradient tests.
- Updates acados

### DIFF
--- a/leap_c/examples/cartpole/planner.py
+++ b/leap_c/examples/cartpole/planner.py
@@ -100,5 +100,5 @@ class CartPolePlanner(AcadosPlanner[AcadosDiffMpcCtx]):
             x_threshold=self.cfg.x_threshold,
         )
 
-        diff_mpc = AcadosDiffMpcTorch(ocp, export_directory=export_directory, dtype=cfg.dtype)  # type:ignore
+        diff_mpc = AcadosDiffMpcTorch(ocp, export_directory=export_directory, dtype=self.cfg.dtype)
         super().__init__(param_manager=param_manager, diff_mpc=diff_mpc)
--- a/leap_c/examples/chain/planner.py
+++ b/leap_c/examples/chain/planner.py
@@ -121,6 +121,6 @@ class ChainPlanner(AcadosPlanner[AcadosDiffMpcCtx]):
             ocp,
             initializer=initializer,
             export_directory=export_directory,
-            dtype=cfg.dtype,  # type:ignore
+            dtype=self.cfg.dtype,
         )
         super().__init__(param_manager=param_manager, diff_mpc=diff_mpc)
--- a/leap_c/examples/hvac/planner.py
+++ b/leap_c/examples/hvac/planner.py
@@ -154,8 +154,8 @@ class HvacPlanner(AcadosPlanner[HvacPlannerCtx]):
             diff_mpc_kwargs = {}
 
         diff_mpc = AcadosDiffMpcTorch(
-            ocp, **diff_mpc_kwargs, export_directory=export_directory, dtype=cfg.dtype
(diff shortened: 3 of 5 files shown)


tests/leap_c/ocp/acados/test_acados_torch.py
```diff
@@ -791,3 +781,99 @@ def forward_func(p_global):
...
+
+def test_backward_lmpc(
+    diff_mpc: AcadosDiffMpcTorch,
+    diff_mpc_with_stagewise_varying_params: AcadosDiffMpcTorch,
+    n_batch: int = 4,
+    max_batch_size: int = 10,
+    dtype: torch.dtype = torch.float64,
+    noise_scale: float = 0.1,
+) -> None:
+    """Test backward pass of AcadosDiffMpcTorch (LMPC example) using finite differences.
```
dirkpr: LMPC -> Pointmass. But you are right that it is a linear mpc

tests/leap_c/ocp/acados/test_acados_torch.py
```diff
@@ -673,98 +667,94 @@ def forward_func(p_global):
...
-                print(f"✓ {test_name} gradient check passed")
-            except Exception as e:
-                print(f"✗ {test_name} gradient check failed: {e}")
-                raise
-        print(f"Times taken in seconds for batch_size {n_batch}: {times}.")
+def create_grid(offset, num):
+    return 1e-3 * np.linspace(0, 1, num=num) + offset - 5 * 1e-4
 
 
 # NOTE: Useful for debugging the sensitivities.
```
dirkpr: I think it is better to make a function and document it is for debugging rather than having a lot of commented code in here. Can use a DEBUG flag to show where you used it.

leap_c/autograd/torch.py
```diff
@@ -70,11 +70,11 @@ def _to_np(data):
         return data
 
 
-def _to_tensor(data, device, dtype=torch.float32):
+def _to_tensor(data, device):
```
dirkpr: Not sure if we want to remove this and keep casting to `torch.tensor` (`torch.float32`) on the higher level

tests/leap_c/ocp/acados/test_acados_torch.py
```diff
@@ -205,6 +209,7 @@ def test_backup_functionality(diff_mpc: AcadosDiffMpcTorch) -> None:
         np.testing.assert_allclose(
             solutions[0][i],
             solutions[1][i],
+            rtol=4 * 1e-7,
```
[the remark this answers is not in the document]
Confectio: Because I wanted to keep the tolerance low, and 3*1e-7 was not enough :'D .

leap_c/autograd/torch.py
Confectio: I just talked with Jasper and I will add a flag to somewhere in the DiffMpc construction soon, to allow float64, but make float32 standard. Would that work for you?

PR #215, opened 2026-01-12
Documentation use leap c setup action

### LINKED ISSUE #214: Documentation workflow does not use common `set_up` action
The workflow in `documentatation.yml` uses commands to set up leap-c that are different from the rest, as defined in `actions/setup-leap-c/action.yml`, which means code duplication that should be fixed.

dirkpr:
Resolves #214


JasperHoffmann (approved):
Looks good!

PR #228, opened 2026-01-16
Add Hyperparameter Tuner to scripts

JasperHoffmann:
This `PR` adds a `smac3` rscript to tune the params of a MPC using hyperparameter tuning frameworks. We add dependencies to `pyproject.toml`

We also added torch to the dependencies, as this was missing before.

### DIFF
--- a/scripts/run_smac_tuning.py
+++ b/scripts/run_smac_tuning.py
@@ -29,7 +29,7 @@ from leap_c.run import default_controller_code_path, default_output_path
 
 @dataclass
 class SmacTuningConfig:
-    """Configuration for SMAC-based hyperparameter tuning.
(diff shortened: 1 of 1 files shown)


Confectio (approved):
Looks good to me :). 
One comment:
We deliberately didnt specify torch in the dependencies, since we explicitly mention in the installation instructions a step to get the torch version people want (e.g., CPU/GPU torch). Also, Numpy and casadi are kind of unnecessary, since they come with getting acados anyways (though I dont mind leaving the packages in the dependencies like this, should not do any harm).

JasperHoffmann:
I think for now I prefer that we explicitely state the torch dependency :) Users could pre-install there according torch version beforehand in the env and we will not update it. But we should mention this in the docs.

PR #237, opened 2026-01-19
Performance improvements

FilippoAiraldi:
With this PR, I am collecting some changes that I made since working on `leap-c` aimed at improving performance. What follows is a description of each commit. I divided the two commits in two separate groups: commits that modify code that runs only once or twice (e.g., instantiating a controller), and commits modifying hot code that lives inside the simulation loop. Obviously, the latter should have larger impact on the overall runtime.

# Hot code commits

- _**https://github.com/leap-c/leap-c/pull/237/changes/99d653451c021dcb721d74b53e51a6f97ca7538b (using `inference_mode` instead of `no_grad` during inference (should be faster).)**_ replaces in each trainer's `act` method the `torch.no_grad` with `torch.inference_mode` which should be faster and cheaper to compute.
- _**https://github.com/leap-c/leap-c/pull/237/commits/80317b55dffa2ebe62fa12084137ce88b672b686 (avoiding useless copy of data during type conversion)**_ avoids creating deep copies of tensors/arrays during type conversion, if already with the correct type and on the specified device.
- _**https://github.com/leap-c/leap-c/pull/237/changes/da5302630feb340f68025434080fe3519506a246 (removed as many keyword args as possible in tight forward/backward loop of AcadosDiffMpcFunction)**_ removes keyword arguments in favor of positional arguments in many of the calls inside `AcadosDiffMpcFunction.foward` and `AcadosDiffMpcFunction.backward`. Calls with keywords arguments are notoriously slower than with positional (not by much, but it's a low hanging fruit).
- _**https://github.com/leap-c/leap-c/pull/237/changes/a35054e1c31303c31fb1f9d105689a4562eb44d5 (reduced indexing and retrieval of bounds in prepare_batch_solver)**_ removes repeated indexing of arrays with the same index and repeated retrieval of `ocp`'s bound properties.
- _**https://github.com/leap-c/leap-c/pull/237/changes/ee785396a8a162bc5e589c9a4e9cd28407cc7eda (caching created seed_vec for consecutive calls to AcadosDiffMpcFunction.sensitivity)**_ caches the created `seed_x` and `seed_u`. In this way, we sacrifice a bit of memory but in turn we don't need to re-create these `list[tuple[int, array]]` from scratch at every call, if already created once.  
- _**https://github.com/leap-c/leap-c/pull/237/changes/39a815e2a2ea9edc6467c1e25c68d3ddd00e374b (removed local methods inside AcadosDiffMpcFunction.backward to avoid having to create these closures at every call)**_ turns the methods `_adjoint`, `_jacobian` and `_safe_sum` into class methods. In this way, these closures don't have to be created anew everytime `AcadosDiffMpcFunction.backward` is called.
- _**https://github.com/leap-c/leap-c/pull/237/changes/39c02b9fb1529449d9a361efe4fb327f9bb19320 (disabling sanity checks after first call)**_ disables `acados`' `eval_adjoint_solution_sensitivity` sanity checks after they have been run the first time. Since these checks probe only the sizes of the OCP, I believe it suffices to check them only once (as long as the OCP does not get re-compiled afterwards; correct me if I am wrong). With this commit, we avoid unnecessary overhead due to these checks in successive calls. 
NOTE: I have not disabled the sanity checks at the bottom of `_adjoint`. Since this call to `eval_adjoint_solution_sensitivity` is data-dependent, it makes sense to leave the checks on such that they can always be run on each new, possibly different, input. 

# Cold(er) code commits

- _**https://github.com/leap-c/leap-c/pull/237/changes/9c946b1fb07b789de57bedaf85c821d5f1299d74 (removed iteration over small lists)**_ just removes some `not in` iterating over lists of one or two elements in favor of direct comparison with said elements (avoids constructing these lists).
- _**https://github.com/leap-c/leap-c/pull/237/changes/817bec9c9dfa7a0418f9a8f64b5de0183376282e (reduced referencing to ocp.solver_options and ocp.model)**_ is more of a style choice, but it also reduces the amount `ocp` properties are fetched, so the code should be ever so slightly faster.
- _**https://github.com/leap-c/leap-c/pull/237/changes/7a56ce90d03c02addb32d83136fd5a035b4c8fd0 (minor refactoring)**_ improves some docstrings and swaps the large if-else sensitivity statements in favor of a (hopefully, cleaner) match-case.

### DIFF
--- a/leap_c/ocp/acados/parameters.py
+++ b/leap_c/ocp/acados/parameters.py
@@ -262,12 +262,13 @@ class AcadosParameterManager:
 
         batch_size = inferred_batch_size if inferred_batch_size is not None else batch_size or 1
 
-        # Get default parameter array and tile for batch
+        # Get default parameter array and replicate it along the batch dimension - if no overwrites
+        # are passed, just return a broadcasted view to avoid unnecessary memory allocation;
+        # otherwise, create a tiled array (is writeable, so overwrites can be applied afterwards)
         default_flat = self.learnable_parameters_default.cat.full().flatten()
-        batch_param = np.tile(default_flat, (batch_size, 1))
-
         if not overwrites:
-            return batch_param
+            return np.broadcast_to(default_flat, (batch_size, default_flat.size))
+        batch_param = np.tile(default_flat, (batch_size, 1))
 
         # Build index mappings for efficient vectorized assignment
         for param_name, values in overwrites.items():
@@ -373,13 +374,16 @@ class AcadosParameterManager:
         # Infer batch size from overwrite if not provided.
         # Resolve to 1 if empty, will result in one batch sample of default values.
         batch_size = next(iter(overwrite.values())).shape[0] if overwrite else batch_size or 1
-
-        # Create a batch of parameter values
         Np1 = self.N_horizon + 1
-        batch_parameter_values = np.tile(
-            self.non_learnable_parameters_default.cat.full().reshape(1, -1),
-            (batch_size, Np1, 1),
-        )
+        expected_shape = (batch_size, Np1, self.non_learnable_parameters.shape[0])
+
+        # Create a batch of parameter values - if indicators are not needed and no overwrites are
+        # passed, just return a broadcasted view to avoid unnecessary memory allocation; otherwise,
+        # create a tiled array (is writeable, so indicators and overwrites can be applied afterward)
+        nonlearn_param_default_flat = self.non_learnable_parameters_default.cat.full().reshape(-1)
+        if not (self.need_indicator or overwrite):
+            return np.broadcast_to(nonlearn_param_default_flat, expected_shape)
+        batch_parameter_values = np.tile(nonlearn_param_default_flat, (batch_size, Np1, 1))
 
         # Set indicator for each stage
         if self.need_indicator:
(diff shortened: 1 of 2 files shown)


leap_c/ocp/acados/diff_mpc.py
```diff
@@ -348,67 +307,125 @@ def sensitivity(
...
+                raise ValueError(f"Unexpected `field_name` {field_name} encountered.")
 
         setattr(ctx, field_name, sens)
-
         return sens
+
+    def _adjoint(
+        self, x_seed: np.ndarray | None, u_seed: np.ndarray | None, with_respect_to: str
+    ) -> np.ndarray | None:
+        """Compute the adjoint sensitivity via backpropagation."""
```
JasperHoffmann: Do you how more expensive this part is, I personally prefer if they are just local functions stylewise :) But can also live with them being methods of the class!

leap_c/ocp/acados/parameters.py
```diff
@@ -373,13 +374,16 @@ def combine_non_learnable_parameter_values(
...
-        )
+        expected_shape = (batch_size, Np1, self.non_learnable_parameters.shape[0])
+
+        # Create a batch of parameter values - if indicators are not needed and no overwrites are
+        # passed, just return a broadcasted view to avoid unnecessary memory allocation; otherwise,
+        # create a tiled array (is writeable, so indicators and overwrites can be applied afterward)
+        nonlearn_param_default_flat = self.non_learnable_parameters_default.cat.full().reshape(-1)
+        if not (self.need_indicator or overwrite):
+            return np.broadcast_to(nonlearn_param_default_flat, expected_shape)
+        batch_parameter_values = np.tile(nonlearn_param_default_flat, (batch_size, Np1, 1))
```
JasperHoffmann: NIce :)

JasperHoffmann (approved):
Looks good! Nice that we use the sanity check features :)

PR #238, opened 2026-01-20
Remove QP default warmstarting options

Confectio:
This PR removes the default options regarding QP warmstarting. This 
1. gives the user more flexibility for choosing the desired QP warmstarting options themselves
2. removes the annoying warnings

If you are using an Interior Point QP solver (like HPIPM) this shouldn't make a noticeable difference.

Thanks @josipkh for bringing this to my attention 👍 .


dirkpr (approved):
Looks good!

PR #240, opened 2026-01-20
Fixed log interval in sac_zop

JasperHoffmann:
Fixed a wrong default setting.


dirkpr (approved):
Thanks for fixing this!

PR #248, opened 2026-01-28
Bounded Beta fix on anchor + improvements

FilippoAiraldi:
While perusing the `ModeConcetrationBeta` bounded distribution, I noticed that the current implementation of how the Beta distribution was anchored around the provided anchor point is incorrect. This PR fixes that (details are provided below). On top of that, I freshened up the docstrings, added some extra checks, taken care of the TODOs, etc.

- _**https://github.com/leap-c/leap-c/commit/52584d48376f076c422f37b51f04b30d1978bc5b (mapping anchor to logit (same space as mode))**_: the current implementation appears to take as input the `mode`, which is provided by an `Mlp` and lies in the logit space (i.e., $[-\infty,\infty]$, or what you get via the inverse of the sigmoid) and directly sums it with the inverse of the anchor, that is `anchored_mode = mode + self.inverse(anchor)`. This leads to the Beta being anchored to a wrong anchor: the reason is that `self.inverse(anchor)` lies in the interval $[0,1]$, not $[-\infty,\infty]$, so we are de facto mixing two quantities that are not comparable. The fix I propose is simply to apply the logit function to `self.inverse(anchor)` so as to map it to $[-\infty,\infty]$ before summing it with `mode`. 
I made a [Desmos graph](https://www.desmos.com/calculator/vthvozheik) showing a comparison of the two methods (current and proposed). It depicts in purple the original (not anchored) Beta distribution, in red the current anchored implementation, and in green the proposed one; more information on the notes there.
NOTE: the original docstring was suggesting that `mode` is already provided in the $[0,1]$ space. However, after that, the original code applies `torch.sigmoid(mode)`. This either indicates that `mode` is not actually provided in $[0,1]$, or is a massive bug because we are applying the sigmoid twice to the mode. 

- _**https://github.com/leap-c/leap-c/commit/2d17ef907f519d23d2d09439f9975db8d2728ef5 (better description of log_conc argument)**_ improves the docstring for the `log_conc` argument and renames it to `logit_log_conc`. Since this quantity goes through a sigmoid and only then is exponentiated, in reality it should be regarded as the logit of the logarithm of the concentration.

- _**https://github.com/leap-c/leap-c/commit/2271f1cc5f3cdf3a18895a5f9751fa0e46b4117e (removing unused attributes)**_ removes a couple of unused attributes that I found.

- _**https://github.com/leap-c/leap-c/commit/b4736168a65b1df18919c51f32027f28efa34550 (updated docstrings and few type hints)**_ freshens the docstrings and type hints.

- _**https://github.com/leap-c/leap-c/commit/2df103f1a8f9068e40b1e1dc1814f6b664f35450 (minor refactoring)**_ applies some changes, but no functionality is changed.

- _**https://github.com/leap-c/leap-c/commit/b1751c4e2fee4a56dda0e5a80d8d36d63fa05ea7 (removed unnecessary broadcasting)**_ removes unnecessary `[None, :]` broadcasting as it already occurs by default.

- _**https://github.com/leap-c/leap-c/commit/6f060e93307a7214ca8b07e030ec7d66c72277b3 (added exception if provided Box space is unbounded)**_ checks that the space provided to the `BoundedDistribution.__init__` is actually bounded; otherwise, it raises an exception.

- _**https://github.com/leap-c/leap-c/commit/03d645acce8bc9075afd1e6d6d01144ef915d226 (taken care of TODOs asking to check anchor is within space)**_ implements the two TODOs asking to check that the provided anchor does not violate the given bounds, via two assertions (I rebased this commit improperly, a piece of it has been leaked into another commit, ops).

- _**https://github.com/leap-c/leap-c/commit/3679cb4569701c9eae12cd55b0cfd096e93ca739 (added tests for ModeConcentrationBeta)**_ implements the missing tests for `ModeConcentrationBeta`.

- _**https://github.com/leap-c/leap-c/commit/149ef83f0d6e2214fd55753b3811e469bc86ada5 (implemented anchoring for ScaledBeta via mode)**_ takes care of the last TODO, i.e., implementing anchoring for `ScaledBeta`. The way is implemented is similar to `ModeConcentrationBeta`: compute the mode from `alpha` and `beta`, and map `mode` and `anchor` to the same logit space; sum them together and lift them up back to the $[0,1]$ space; finally, recover the new `alpha` and `beta` shapes. I also implemented the missing test for anchoring in `ScaledBeta`.

### DIFF
--- a/leap_c/torch/nn/bounded_distributions.py
+++ b/leap_c/torch/nn/bounded_distributions.py
@@ -243,6 +243,8 @@ class ScaledBeta(BoundedDistribution):
         log_beta_min: The minimum value for the logarithm of the beta parameter.
         log_alpha_max: The maximum value for the logarithm of the alpha parameter.
         log_beta_max: The maximum value for the logarithm of the beta parameter.
+        padding: The amount of padding to distance the action of the bounds, when using the inverse
+            transformation for the anchoring. This improves numerical stability.
     """
 
     scale: torch.Tensor
@@ -251,6 +253,7 @@ class ScaledBeta(BoundedDistribution):
     log_beta_min: float
     log_alpha_max: float
     log_beta_max: float
+    padding: float
 
     def __init__(
         self,
@@ -259,6 +262,7 @@ class ScaledBeta(BoundedDistribution):
         log_beta_min: float = -10.0,
         log_alpha_max: float = 10.0,
         log_beta_max: float = 10.0,
+        padding: float = 1e-4,
     ) -> None:
         """Initializes the `ScaledBeta` module.
 
@@ -268,6 +272,8 @@ class ScaledBeta(BoundedDistribution):
             log_beta_min: The minimum value for the logarithm of the beta parameter.
             log_alpha_max: The maximum value for the logarithm of the alpha parameter.
             log_beta_max: The maximum value for the logarithm of the beta parameter.
+            padding: The amount of padding to distance the action of the bounds, when using the
+                inverse transformation for the anchoring. This improves numerical stability.
 
         Raises:
             ValueError: If the provided space is not bounded.
@@ -277,6 +283,7 @@ class ScaledBeta(BoundedDistribution):
         self.log_beta_max = log_beta_max
         self.log_alpha_min = log_alpha_min
         self.log_beta_min = log_beta_min
+        self.padding = padding
 
         self.register_buffer("loc", torch.tensor(space.low, dtype=torch.float32))
         self.register_buffer("scale", torch.tensor(space.high - space.low, dtype=torch.float32))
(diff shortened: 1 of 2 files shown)


JasperHoffmann (approved):
Thank you a lot :) So far I have not played around with the Beta distribution, but now I want to test it!

JasperHoffmann:
> I have a curiosity. Why is residual learning not allowed in action noise regimes? From a code perspective, it should be easy to implement, and the gradients should flow as well. I do not see any hinder to it. Am I missing something?

Currently, I deactivated some parts of it just to keep things simple as I did not try out all the coding paths yet. Probably the implementation of the MPCActor together with the BoundedDistribution can be improved. Currently, we are abusing the SquashedGaussian too much :D Maybe the `HierachicalMPCActor` should be refactored.

For the `BetaDistribution` I really want to try it out soon and compare!

PR #250, opened 2026-02-03
Numerical improvements and more tests for bounded distributions

FilippoAiraldi:
I was trying to add tests on the log probabilities for the various distributions, when I noticed that some numerical improvements or simplifications could be enjoyed by the bounded distributions. I describe below what each commit does. As usual, if you don't approve of a change, please let me know!

* _**https://github.com/leap-c/leap-c/commit/ad84e8e0e9f1f008a49e8129e22223d45141e4fb (fixing bug: not all bounded distributions support log_std in forward)**_ removes the args `log_std=None` passed to `self.param_distribution` in `HierachicalMPCActor` and swaps it for `deterministic=True`; this enables the use of any distribution, not just `SquashedGaussian`.

* _**https://github.com/leap-c/leap-c/commit/9a0172d3b81455e6564e36d25e37ce78a93695ef (better type hinting, docstrings, and removing numpy dependency)**_ improves the docstrings here and there as well as the type hints, and removes avoidable calls to NumPy (also, ever so slightly slower than calling the `math` lib on a scalar, as far as I know)

* _**https://github.com/leap-c/leap-c/commit/6e13ae08c8bd5e59338710e0f865d2330eb25015 (added all attributes as buffers (better for saving and loading); enabled these attributes to be array-like also)**_ moves all numerical attributes of bounded distributions to buffers (e.g., paddings and min/max bounds). This enables better saving and loading of PyTorch states, if the need ever arises. Also, these arguments have now been made array-like so that, e.g., different paddings can be specified for different dimensions. 

* _**https://github.com/leap-c/leap-c/commit/85854f0199230c4509f56a7c6dc569b979186633 (refactored tests for bounded distrs: now they all test for deterministic and not deterministic behavior)**_ just refactors all tests to use `pytest.mark.parametrize` to test both for deterministic and stochastic case, and makes all tests random (e.g., bounded space and bounded distribution parameters are now randomly generated). This is merely a change to reduce code duplication and increase coverage; the functionality of the tests was left untouched.

* _**https://github.com/leap-c/leap-c/commit/e8e9109559fb288ddaf1e7efec4906ff028d33d3 (refactored ScaledBeta to avoid numerical issues and use more efficient computations; added tests)**_ refactors `ScaledBeta` to make sure that concativity is always guaranteed (adds a small `torch.finfo(...).eps` to make sure of this); uses slightly more efficient computations (e.g., `torch.addcmul`; simplifies `beta` when `anchor` is provided); avoids in-place operations with `log_prob` in case of shape mismatch; adds test for log_prob computations.

* _**https://github.com/leap-c/leap-c/commit/b4355405476ac1cdb3b9d0609a70578b88ec635e (streamlined ModeConcentrationBeta and also here added numerical stability improvements; added tests)**_ is similar to above, but for `ModeConcentrationBeta`. Ensures `concentration > 2` to avoid division by zero; removes `self._beta_dist` and `self._update_distribution` (I guess these were vibe-coding remnants :D); adds test for log_prob computations.

* _**https://github.com/leap-c/leap-c/commit/b6c30203479520b4016f9aa71f71f20fad17ad68 (same for SquashedGaussian; numerical stab improvements and more tests)**_ is, again, similar to the two above, for `SquashedGaussian`. In particular, it uses [this trick](https://github.com/tensorflow/probability/blob/main/tensorflow_probability/python/bijectors/tanh.py#L69-L80) to avoid numerical issues in `log_prob` and to get rid of the hard-coded `1e-6` offset that was previously used.
NOTE: removing the `1e-6` offset that was previously used in `log_prob` might change quite a lot the expected `log_prob`. In particular, this offset was quite visible for high `log_prob` (when comparing them to their expected values), and it was testing these strange log probabilities that motivated this PR. If we want to PRESERVE this behaviour, then the numerical trick I propose in this last commit cannot be applied. However, this numerical trick should compensate and take care of the reason why the `1e-6` was added in the first place.

---

Tests failed, so I added the following two commits

* _**https://github.com/leap-c/leap-c/pull/250/changes/21404750c31ecf70a6af183d2a20b22b9e1f7009 (use self.padding instead of torch.finfo.eps; set zero padding in tests to avoid distortion)**_ swaps `torch.finfo.eps` in favour of `SquashedGaussian.padding`; sets the padding to zero during these tests, otherwise failures occur due to distortions due to eps.
* _**https://github.com/leap-c/leap-c/pull/250/changes/4c83f89971972bb2ca2a53fdc082e4cbc1b80a65 (add `HierachicalMPCActorConfig.distribution_kwargs` to allow specifying, among others, padding of distribution)**_ takes care of another two tests failing because of padding distortions; so, I added a config entry for the MPC actor config to specify, among others, padding arguments and set it to zero for the tests.

---

After some commits tweaking tests, minor bugs and improvements, I added the following commit

* _**https://github.com/leap-c/leap-c/pull/250/changes/a37b6d42949040e45cfa25707d106cdbd0feccd1 (added support for sample_shape in bounded distributions)**_ adds `sample_shape` to the forward method of `BoundedDistribution`. This allows to draw more than one sample per batched distribution, so that the samples are now `(*sample_shape, n_batch, dim_space)`. This is especially useful when many parameters from the same distribution are needed (e.g., RQMC).

### DIFF
--- a/tests/leap_c/torch/nn/test_bounded_dist.py
+++ b/tests/leap_c/torch/nn/test_bounded_dist.py
@@ -74,11 +74,9 @@ def test_squashed_gaussian_anchor(deterministic: bool, single_sample: bool) -> N
         if deterministic
         else torch.from_numpy(rng.normal(size=shape)).requires_grad_()
     )
-    log_std = (
-        torch.from_numpy(rng.normal(size=shape))
-        .clamp(distribution.log_std_min, distribution.log_std_max)
-        .requires_grad_()
-    )
+    log_std = torch.from_numpy(
+        rng.uniform(distribution.log_std_min, distribution.log_std_max, size=shape)
+    ).requires_grad_()
     anchor = torch.from_numpy(rng.uniform(space.low, space.high, size=shape))
     samples, log_prob, _ = distribution(mean, log_std, deterministic, anchor, sample_shape)
     assert samples.shape == sample_shape + shape
@@ -121,8 +119,8 @@ def test_squashed_gaussian_log_prob(single_sample: bool) -> None:
     # generate random Gaussian parameters and samples with associated log probs
     shape = batch_shape + (event_dim,)
     mean = torch.from_numpy(rng.normal(size=shape))
-    log_std = torch.from_numpy(rng.normal(size=shape)).clamp(
-        distribution.log_std_min, distribution.log_std_max
+    log_std = torch.from_numpy(
+        rng.uniform(distribution.log_std_min, distribution.log_std_max, size=shape)
     )
     samples, log_prob, _ = distribution(mean, log_std, sample_shape=sample_shape)
     log_prob = log_prob.squeeze(-1)


leap_c/torch/nn/bounded_distributions.py
```diff
@@ -378,17 +426,16 @@ class ModeConcentrationBeta(BoundedDistribution):
...
+    log_conc_max: torch.Tensor
+    padding: torch.Tensor
 
     def __init__(
         self,
         space: Box,
-        log_conc_min: float = np.log(2.0),
-        log_conc_max: float = np.log(100.0),
-        padding: float = 1e-4,
+        log_conc_min: ArrayLike | float = log(2.0),
```
dirkpr: Should we just return an array with one element here? I remember we had this topic at a recent PR in the context of distribution-defining parameters.

leap_c/torch/nn/bounded_distributions.py
```diff
@@ -444,62 +506,41 @@ def forward(
...
+        concentration = log_conc.exp()
 
+        # generate sample - if deterministic, use mode directly
         if deterministic:
-            # Deterministic: just use the mode scaled to action space
-            y = mode
-            log_prob = torch.zeros_like(mode)
+            out_shape = sample_shape + torch.broadcast_shapes(mode.shape, self.loc.shape)
+            y = mode.broadcast_to(out_shape)
+            log_prob = mode.new_zeros(()).broadcast_to(out_shape)
```
dirkpr: As above, use `y.new_zeroes(())` to avoid broadcast

leap_c/torch/nn/bounded_distributions.py
```diff
@@ -180,32 +212,32 @@ def forward(
...
-        if std is not None:
-            log_prob = -0.5 * ((y - mean) / std).square() - log_std - 0.5 * np.log(2 * np.pi)
+            mean = mean + inv_anchor
+
+        # sample and compute log probability - if deterministic, use the mean and assign 0 log_prob
+        stats = {}
+        if deterministic or log_std is None:
+            out_shape = sample_shape + torch.broadcast_shapes(mean.shape, self.loc.shape)
+            y = mean.broadcast_to(out_shape)
+            log_prob = mean.new_zeros(()).broadcast_to(out_shape)
```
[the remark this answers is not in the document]
FilippoAiraldi: Not sure I get the comment. The idea is to create a tensor with a single `0` (i.e., with shape `()`) and then broadcast it to `out_shape` to avoid allocating a whole tensor of zeros with the same shape. `y.new_zeros` achieves exactly the same as `mean.new_zeros`, still requiring `broadcast_to`.

leap_c/torch/nn/bounded_distributions.py
FilippoAiraldi: Looking at the definition of [ArrayLike](https://numpy.org/devdocs/reference/typing.html#numpy.typing.ArrayLike), you are correct, `ArrayLike` already encompasses scalar values. I will remove the union with `float` where redundant.

Issue #244, opened 2026-01-22
Running CI tests in forks


josipkh:
It seems like the CI tests are not run when pushing to branches different than "main" in forks.
Perhaps it would be good to have this option, instead of first opening a PR and addressing the failing tests afterwards.

Is there a reason to skip running CI tests in forks?

PR #252, opened 2026-02-04
Run CI in forks

josipkh:
addresses #244


dirkpr (approved):
Thanks! Looks good to me to always run linting and tests.

PR #258, opened 2026-02-11
Discount factor not trickled down to batch solver

FilippoAiraldi:
# Bug Description

All leap-c MPC-RL methods have a discount factor in their script, e.g.,
https://github.com/leap-c/leap-c/blob/1cc842267f8847dd354845fa9d39456ad81d85d2/scripts/run_sac_fop.py#L61

However, I found out that this discount factor is never properly passed to the constructor of `AcadosDiffMpcFunction`
https://github.com/leap-c/leap-c/blob/1cc842267f8847dd354845fa9d39456ad81d85d2/leap_c/ocp/acados/diff_mpc.py#L123-L132

and so `_set_discount_factor` is never called down the line. De facto, this means that, no matter what `cfg.trainer.gamma` is specified, we always end up using acados's default [`cost_scaling`](https://docs.acados.org/python_interface/#acados_template.acados_ocp_options.AcadosOcpOptions.cost_scaling) in our MPC problems. 

# Fix

Assuming that this is not intended behaviour and we instead would like `AcadosDiffMpcFunction` to be instantiated with the discount factor specified in `cfg.trainer.gamma`, this PR adds, alongside `dtype`, the `discount_factor` field (as well as the missing `n_batch_max` and `num_threads_batch_solver`) to all the examples' configurations. It then passes `cfg.trainer.gamma` to `create_controller` so that the discount factor in the config will be set to this value.

Let me know if I misinterpreted and the use of the default `None` discount factor was indeed intended behaviour!

PS: I have also streamlined `examples/__init__.py` by removing some code duplication.

PPS: I removed the older `n_batch_max` in favour of the newer `n_batch_init`; both were still in use (for backward compatibility? But the argument was never set externally anyway)

### DIFF
--- a/.github/actions/setup-leap-c/action.yml
+++ b/.github/actions/setup-leap-c/action.yml
@@ -15,7 +15,7 @@ runs:
         path: ${{github.workspace}}/leap_c/external/acados
         repository: acados/acados
         github-token: ${{ github.token }}
-        run-id: 21400273673
+        run-id: 22190081244
 
     - name: Rename artifact directories
       working-directory: ${{github.workspace}}/leap_c/external/acados
--- a/.github/workflows/run_ruff.yml
+++ b/.github/workflows/run_ruff.yml
@@ -1,9 +1,8 @@
 name: Ruff Linting and Formatting
 on:
   push:
-    branches: [ "main" ]
   pull_request:
-    branches: [ "main" ]
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,9 +2,7 @@ name: Tests
 
 on:
   push:
-    branches: [ "main" ]
   pull_request:
-    branches: [ "main" ]
 
 jobs:
 
@@ -47,9 +45,12 @@ jobs:
         pytest tests/leap_c/test_scripts.py -v -s --cov=leap_c --cov-report=xml:coverage-${{ matrix.test-group.env }}.xml --cov-report=term
 
     - name: Upload coverage to Codecov
+      env:
+        CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+      if: env.CODECOV_TOKEN != ''
       uses: codecov/codecov-action@v4
       with:
-        token: ${{ secrets.CODECOV_TOKEN }}
+        token: ${{ env.CODECOV_TOKEN }}
         name: ${{ matrix.test-group.env }}-${{ matrix.test-group.script }}
         flags: integration-tests
 
@@ -80,8 +81,11 @@ jobs:
         pytest tests/leap_c/ -v --ignore=tests/leap_c/test_scripts.py --cov=leap_c --cov-report=xml:coverage-unit.xml --cov-report=term
 
     - name: Upload coverage to Codecov
+      env:
+        CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+      if: env.CODECOV_TOKEN != ''
       uses: codecov/codecov-action@v4
       with:
-        token: ${{ secrets.CODECOV_TOKEN }}
+        token: ${{ env.CODECOV_TOKEN }}
         name: unit-tests-${{ matrix.python-version }}
         flags: unit-tests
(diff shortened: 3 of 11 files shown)


FreyJo (approved):
Looks good to me!
Setting the discount factor in `leap_c/ocp/acados/utils/create_solver.py` could be simplified, as we made it can now (since ~1year in acados) be done via `acados_ocp.solver_options.cost_scaling`. So it can be done before solver creation.

JasperHoffmann:
Maybe before this is merged, this is not a bug. I think only in MPC as a critic this should be actually be used.

leap_c/examples/__init__.py
```diff
@@ -88,73 +89,72 @@ def create_env(env_name: ExampleEnvName, **kw: Any) -> Env:
...
+# controllers are a superset of planners
+ExampleControllerName: TypeAlias = ExamplePlannerName
+
+
+def _create_from_registry(
+    kind: Literal["planner", "controller"],
+    name: str,
+    reuse_code_base_dir: Path | None,
+    **kwargs: Any,
+) -> ParameterizedPlanner[CtxType] | ParameterizedController[CtxType]:
```
JasperHoffmann: looks good!

leap_c/examples/hvac/planner.py
```diff
@@ -60,14 +60,24 @@ class HvacPlannerConfig:
         N_horizon: The number of steps in the MPC horizon
             (default: 96, i.e., 24 hours in 15-minute steps).
         param_interface: Determines the exposed parameter interface of the planner.
-        param_granularity: Determines the granularity of the parameters
+        param_granularity: Determines the granularity of the parameters.
+        discount_factor: discount factor along the MPC horizon.
```
JasperHoffmann: Personally, I am not sure whether we need that many options here. Let's quickly discuss the separate points:
1. `n_batch_init` I think you will never access this argument anyways here. I think lazy creation is enough. Do you agree?
2. `num_threads_batch_solver` We internally had the discussion before, that we allow to set the num threads via a global method, like this style and provide a good default value. https://docs.pytorch.org/docs/stable/generated/torch.set_num_threads.html
3. `discount_factor` As discussed below, I am currently not aware why one should introduce a discount factor in general to the planner. However, currently the abstract planner also provides `Q-values` already. Thus, I would suggest that we indeed expose the `gamma` in the creation :) But just not set it per default the same as the RL agents one. We could maybe make this explicit in the args docstirng. Let me know what you think :)

scripts/run_sac_fop.py
```diff
@@ -136,7 +136,9 @@ def run_sac_fop(
     trainer = SacFopTrainer(
         val_env=val_env,
         train_env=create_env(cfg.env),
-        controller=create_controller(cfg.controller, reuse_code_dir),
+        controller=create_controller(
+            cfg.controller, reuse_code_dir, discount_factor=cfg.trainer.gamma
```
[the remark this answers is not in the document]
FilippoAiraldi: Perfect, I'll remove the discount from SAC-ZOP and SAC-FOP

FilippoAiraldi:
Update: merged with changes from #259, and removed passing $\gamma$ to SAC-ZOP and SAC-FOP's controllers as discussed.

leap_c/examples/hvac/planner.py
JasperHoffmann: 4. With regards to 4. Sounds nice :) Looking forward to the PR, even though from my intuition this should be maybe all handled in the trainer. Maybe the `AcadosDiffMpc` just should adapt dynamically to the inputted precision.  But I think there are many viable options here.

3. I think the problem is that $\gamma^n$ might be wrong if the time discretization is not aligned. But anyways for now I think we should just set it to None and then use the default provided :) So your approach should be fine.

2. Yes, I was suggesting that:
```python
def set_default_num_threads_batch_solver(num_threads: int): ...
    # overwrite DEFAULT here

class AcadosDiffMpcFunction:
    def __init__(self, ..., num_threads: int | None = None):
         self.num_threads = ...
         
class PlannerCfg:
    num_acados_threads: int | None = None
```
Then you could use `set_default_num_threads_batch_solver`.

@dirkpr What is your final verdict? I can live with both options.

Issue #263, opened 2026-02-20
Code reuse still not working


FilippoAiraldi:
Updated leap-c and acados after PR #262, but code reuse is still not working. Running 

```bash
python scripts/run_sac_fop.py --env=cartpole --reuse_code --with-val --variant=fop
```

the second time I still get the message

```bash
OCP formulation has changed, code reuse not possible.
List of mismatching fields:
 ['dims.n_global_data']
```

After some digging, it seems that even though `"n_global_data"` had been added to the set of JSON fields to be ignored, it was not properly removed (because it lives nested under another dict, `"dims"`). I made a PR in acados to overcome this issue (https://github.com/acados/acados/pull/1788).

PR #264, opened 2026-02-20
Bumping acados version for code reuse fix

FilippoAiraldi:
Bumping acados version (with new fix for code reuse) as per issue #263.


dirkpr (approved):
Looks good!

PR #270, opened 2026-03-05
Fix pyproj installation issues

ArgoJ:
Consistency issues and other setuptools setup such that leap-c is correclty installed and find all subfolder (which it didn't before)


dirkpr (changes requested):
Needs some clarification on why `ensure_external_cost_yrefs` is necessary. I expect the `pyproject.toml` changes to be breaking the installation.

ArgoJ:
Servus,
The leap_c is a faulty try that I had in, sorry I will take that out. Nevertheless, the rest of the changed things in pyproj should work correctly with this fix. Before there were no modules, no subfolders, just the files in the main folder when it was installed.
Regarding the other change using external I have to look it up again. Currently I have no time, therefore this needs to wait a week.
Best regards 
Josua

PR #273, opened 2026-03-18
Hvac Improvements

JasperHoffmann:
## HVAC environment improvements

> **Note:** This PR depends on #272  and should be merged after it.

Refactors and extends the HVAC RL environment: updates house parameters to
Freiburg climate, introduces dict observations, adds weekday occupancy gains,
improves dataset/episode length defaults, makes dynamics CasADi SX-friendly,
adds render info for dynamics params, refines default settings and comfort
bounds, updates the parameter interface, and renames the temperature cost
weight to log-space (`log_q_Ti`). Adds HVAC training scripts and tests.

### Commits
- [e1513a1](https://github.com/leap-c/leap-c/pull/273/commits/e1513a1632d127e750d8f7bbf8bc3f65f2f149fb)  feat(hvac): update house parameters to Freiburg
- [0084084](https://github.com/leap-c/leap-c/pull/273/commits/008408488be1ef352a4193df03b9a79f05da07ed) feat(hvac): update extractor with instance normalization
- [dea4cd2](https://github.com/leap-c/leap-c/pull/273/commits/dea4cd22c425a32bdeab98ef8841c610f4fe983a) feat(hvac): made dynamics SX friendly
- [d47cf42](https://github.com/leap-c/leap-c/pull/273/commits/d47cf42b092c1e45d9a90729ed0739c21f758695) feat(hvac): introduce dict obs
- [af1f02a](https://github.com/leap-c/leap-c/pull/273/commits/af1f02ae539e0a351db7dfa4da32b8404060584b) feat(hvac): add during the week gains to env
- [687b315](https://github.com/leap-c/leap-c/pull/273/commits/687b315f2c887d4547fcf059e20ae99e7f1a573c) feat(hvac): small fixes to dataset, min_price and episode length
- [a0aa146](https://github.com/leap-c/leap-c/pull/273/commits/a0aa146bfb31fd153aca9d1469c49f85957cb7e3) feat(hvac): Add dynamics param to render info
- [e06e8f6](https://github.com/leap-c/leap-c/pull/273/commits/e06e8f6d48b5075a0c45ede1b4b2f5830ba14ece) refactor(hvac): updated default settings of env
-  [626604a](https://github.com/leap-c/leap-c/pull/273/commits/626604a4a70dd97d69a5534ade48de607b648042) refactor(hvac): updated comfort temperature bound
[...]

PR #274, opened 2026-03-18
CrossQ

JasperHoffmann:
## CrossQ algorithm with ZOP variant

> **Note:** This PR depends on #273  and should be merged after it.

Adds the CrossQ off-policy RL algorithm and its ZOP (zeroth-order policy)
variant. Introduces a batchnorm option to the MLP module (required by CrossQ's
critic architecture) and finalises the CrossQ training scripts for HVAC.
### Commits
- feat(mlp): add batchnorm option
- feat(CrossQ): add CrossQ
- feat(CrossQ): add CrossQ for ZOP
- feat(Cross-Q): finalized cross-q scripts and trainer
- refactor(Cross-Q): cleaned crossq.py
### Files changed
- `leap_c/torch/rl/crossq.py`
- `leap_c/torch/rl/crossq_zop.py`
- `leap_c/torch/nn/mlp.py`
- `scripts/hvac/run_crossq.py`
- `scripts/hvac/run_crossq_zop.py`

### DIFF
--- a/leap_c/examples/hvac/dataset.py
+++ b/leap_c/examples/hvac/dataset.py
@@ -479,6 +479,9 @@ class HvacDataset:
 
         if split == "test":
             # Return fixed test episodes in order (cycling if needed)
+            if not self._test_indices:
+                # No fixed test episodes configured; fall back to random sampling
+                return self._sample_random(rng, horizon, "all", max_attempts)
             idx = self._test_indices[self._test_episode_counter % len(self._test_indices)]
             self._test_episode_counter += 1
             return idx, max_steps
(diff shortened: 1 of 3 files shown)


leap_c/torch/nn/mlp.py
```diff
@@ -81,11 +81,17 @@ class MlpConfig:
         activation: The activation function to use in the hidden layers.
         weight_init: The weight initialization method to use for the hidden layers. If `None`, no
             initialization will be applied.
+        batchnorm: Whether to use batch normalization between layers.
+
+    Todo (Jasper):
+        Consider switching to RL-specific BatchNorm with stable running average statistics. Current,
```
FilippoAiraldi: Doesn't `BatchNormXd` already use running statistics instead of batch stats by default (`track_running_stats=True`)?

JasperHoffmann: True, I was reading the description here:
https://docs.pytorch.org/rl/0.7/_modules/torchrl/modules/models/batchrenorm.html

Thus, this "more stable" version uses minibatch stats in the beginning of the training and then switches to running stats. Initial experiments did on `hvac` lead to worse results using this implementation.

Will update the Todo description.

Issue #256, opened 2026-02-11
Improved DiffMpc Interface


JasperHoffmann:
Instead of the user managing `p_global` and `p_stagewise` we hide this as an implementation detail.

### Current State
The forward pass of `AcadosDiffMpc` looks like this:

```python
    def forward(  # type: ignore
        self,
        ctx: AcadosDiffMpcCtx | None,
        x0: np.ndarray,
        u0: np.ndarray | None = None,
        p_global: np.ndarray | None = None,
        p_stagewise: np.ndarray | None = None,
        p_stagewise_sparse_idx: np.ndarray | None = None,
    ) -> tuple[AcadosDiffMpcCtx, np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
        """Perform the forward pass by solving the problem instances.
```

Furthermore, we have tight coupling between creating parameters, splitting them into `p_global` and `p_stagewise`, furthermore we have a lot of different objects interacting in non trivial ways: `AcadosParameters`, `AcadosParameterManager`, `AcadosDiffMpcTorch`, and `AcadosDiffMpc`.

### Potential Solution

Hide the parameter management inside `AcadosDiffMpc`! 

Tn the long term the forward pass should look like this:
```python
    def forward(  # type: ignore
        self,
        ctx: AcadosDiffMpcCtx | None,
        x0: np.ndarray,
        u0: np.ndarray | None = None,
        params: dict[str, np.ndarray] = None,
    ) -> tuple[AcadosDiffMpcCtx, np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
        """Perform the forward pass by solving the problem instances.
```
Instead of introducing a separate ParamManager, we introduce a `AcadosDiffOcp` that subclasses `AcadosOcp` to register params.

An example of a future interface, where we just locally register the params could be:
```python
A = ocp.create_SX(
    "A",  # name
    A_MAT_DEFAULT,   # default value
    differentiable=True,    # whether we want to differentiate through it, internally (p_global oder nicht p_global)
    splits="global",  # describes how we should vary the parameters over the horizon.
)

discrete_dyn = A @ ..
```

dirkpr:
> Instead of the user managing `p_global` and `p_stagewise` we hide this as an implementation detail.
> 
....
> An example of a future interface, where we just locally register the params could be:
> 
> A = ocp.create_SX(
>     "A",  # name
>     A_MAT_DEFAULT,   # default value
[...]

PR #288, opened 2026-03-23
Replace double-integrator OCP formulation in HVAC planner

dirkpr:
The previous `HvacPlanner` OCP used the acceleration of heating power (`ddqh`) as the
control input, augmenting the state with `dqh`, `qh`. While this allows to penalise rate changes
and improved solver convergence, it introduced a mismatch between the planner action and the
environment action space. 

# Changes

## OCP reformulation (`acados_ocp.py`, `planner.py`)
- Control input is now `qh` (heating power in kW) directly, removing the double-integrator
  structure and the `qh_prev` state augmentation.
- A quadratic rate-change penalty `q_dqh * (qh - qh_prev)^2` is retained in the cost by
  tracking `qh_prev` as an auxiliary state component, preserving the regularisation benefit
  without the action misalignment.
- Temperature comfort bounds `lb_Ti`/`ub_Ti` are no longer passed as OCP parameters — they had
  no effect there since the OCP model never referenced them. They are now set correctly per-stage
  via `constraints_set("lbx"/"ubx")` on the forward batch solver before each solve.
- `lbu` is set to `[0.0]` to reflect that heating power is non-negative.

## Rendering removed from `StochasticThreeStateRcEnv`
- Matplotlib-based in-loop rendering was removed as it dominated wall-clock time during training
  and validation.

## `step_callback` added to `episode_rollout`
- `episode_rollout` now accepts an optional `step_callback(step, obs, action, reward, info, ctx)`
  invoked after each environment step.
- This decouples post-step logic (logging, rendering, data collection) from the rollout loop
  itself.
- `Trainer` exposes a `_make_val_step_callback()` hook that subclasses can override.

### DIFF
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -1,4 +1,4 @@
-name: Ruff Linting and Formatting
+name: Ruff
 on:
   push:
   pull_request:
--- a/leap_c/examples/hvac/acados_ocp.py
+++ b/leap_c/examples/hvac/acados_ocp.py
@@ -244,13 +244,6 @@ def export_parametric_ocp(
     ocp.cost.cost_type = "EXTERNAL"
     ocp.cost.cost_type_e = "EXTERNAL"
 
-    # Possible functions to allow negative qh but still have meaningful cost.
-    # Both make the solver slow:
-    # smooth approximation of max(0, qh)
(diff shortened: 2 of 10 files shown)


JasperHoffmann (approved):
Looks good! To ensure that our forwad pass in `DiffMpcFunction` stays stateless we should at one point integrate it into the forward pass #256 !

PR #289, opened 2026-03-23
Fix SacFop script and improved CrossQ scripts argparse

JasperHoffmann:
Minor fix to `run_sac_fop.py` and improved argparse of `CrossQ` scripts.

### DIFF
--- a/scripts/hvac/run_crossq.py
+++ b/scripts/hvac/run_crossq.py
@@ -1,12 +1,20 @@
 """Main script to run CrossQ experiments."""
 
-from argparse import ArgumentParser
+from argparse import ArgumentDefaultsHelpFormatter, ArgumentParser
 from dataclasses import asdict, dataclass, field
 from pathlib import Path
-from typing import Literal
+from typing import Literal, get_args
+
+import torch
 
 from leap_c.examples import ExampleEnvName, create_env
-from leap_c.run import default_name, default_output_path, init_run
+from leap_c.run import (
+    default_name,
+    default_output_path,
+    init_run,
+    validate_torch_device_arg,
+    validate_torch_dtype_arg,
+)
 from leap_c.torch.nn.extractor import ExtractorName
 from leap_c.torch.rl.crossq import CrossQTrainer, CrossQTrainerConfig
 
@@ -87,7 +95,7 @@ def create_cfg(
 def run_crossq(
     cfg: RunCrossQConfig,
     output_path: str | Path,
-    device: str = "cuda",
+    device: str,
     with_val: bool = False,
 ) -> float:
     """Run the CrossQ trainer.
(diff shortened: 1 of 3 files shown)


scripts/hvac/run_crossq.py
```diff
@@ -87,7 +93,7 @@ def create_cfg(
 def run_crossq(
     cfg: RunCrossQConfig,
     output_path: str | Path,
-    device: str = "cuda",
+    device: str,
```
FilippoAiraldi, line 96: (also in `run_crossq_zop.py`) hint as `int | str | torch.device`? The same also in `CrossQTrainer.__init__`

scripts/run_sac_fop.py
```diff
@@ -201,6 +201,16 @@ def run_sac_fop(
         help="MPC controller to use as actor. If not provided, it is taken from `--env`.",
     )
     group.add_argument("--with-val", action="store_true", help="Enables validation environment.")
+    group.add_argument(
```
FilippoAiraldi, line 204: good spot! I unintentionally deleted it in the previous PR...

scripts/hvac/run_crossq.py
```diff
@@ -112,29 +118,46 @@ def run_crossq(
...
+    )
+    group = parser.add_argument_group("Run settings")
+    group.add_argument(
+        "--output-path", type=Path, default=None, help="Path to outputs (e.g., logs)."
+    )
+    group.add_argument(
+        "--device", type=validate_torch_device_arg, default="cpu", help="Device to run on."
+    )
+    group.add_argument(
+        "--dtype",
```
FilippoAiraldi, line 132: `--dtype` is added to the arguments but is unused. 

To be in line with SAC and allow users to modify the dtype of the learnable params, maybe we should make the trainers `CrossQTrainer` and `CrossQZop` accept a `dtype` argument and cast the `SacCritic`/`SacActor`/`HierachicalMPCActor`/`nn.Parameter` with `.to(device, dtype)`, as well as pass the dtype to `ReplayBuffer.__init__` (i.e., like `sac_zop.py` does)

FilippoAiraldi (changes requested):
Minor comment on `device`'s type hints, and a major comment for `dtype`. The rest looks good :)

scripts/hvac/run_crossq.py
JasperHoffmann: True probably we should do that for consitency, though I am not sure, wether we want to add this option that prominently. Using float64 on a GPU is orders of magnitudes slower then with float32 :) Are you currently regularly using it?

FilippoAiraldi: Yes, I usually stick with float64. For small feedforward networks, I don't even move them to the GPU as I find the CPU often to be faster...

In any case, at the interface between PyTorch and acados, everything needs to be converted to double (as far as I know, acados does not work with float). It is debatable whether we gain more from using float since it requires this constant conversion from/to double.

PR #290, opened 2026-03-24
Hvac planner absolute power reformulation

dirkpr:
# HVAC: Absolute-value reformulation and dataset simplification

## Motivation

The HVAC planner was previously restricted to non-negative heating power, meaning the system
could only inject heat into the room and not extract it (no cooling). This PR introduces an
LP absolute-value reformulation so that the planner can reason over both positive (heating)
and negative (cooling) power.

A secondary simplification removes the gap-handling logic from `dataset.py`. This logic existed
because only certain months (`valid_months`) were considered valid, requiring the dataset to skip
over gaps. Since we now treat all months as valid (in anticipation of a future environment capable
of cooling), the index advances continuously and the gap-handling code is no longer needed.

Note: the environment itself does not yet model cooling physically. A dedicated environment with
cooling capability will be introduced in a follow-up PR.

### `acados_ocp.py` - QP formulation with absolute-value cost

The OCP was restructured to yield a Quadratic Program with a constant Hessian,
eliminating multiple SQP iterations and the associated `ACADOS_MINSTEP` failures:

- **`qh` promoted from control to state**: The applied heating power `qh` is now part of the
  augmented state `x = [Ti, Th, Te, qh]`. The control input becomes `u = [dqh, t]`, where
  `dqh` is the power increment applied at each step (`qh_new = qh + dqh`).

- **LP absolute-value reformulation**: An auxiliary variable `t >= |qh_new|` is introduced
  via mixed linear constraints `C @ x + D @ u`, enabling the energy cost
  `price * t` to correctly penalise both heating and cooling without introducing
  nonlinearity. The quadratic rate penalty `q_dqh * dqh²` retains a constant Hessian,
  making the full OCP a QP that SQP solves in a single iteration.

### `dataset.py` - Removal of gap-handling logic

With all months treated as valid, the dataset index advances sequentially without jumps.
The following methods and the `valid_months` concept were removed:

- `_find_first_valid_index`
- `_find_next_valid_index`
- `_calculate_steps_until_invalid`
- `valid_months` field and related filtering

This reduces `dataset.py` substantially and makes the dataset logic easier to follow.

### `planner.py` - Corrected action extraction

The action returned to the environment is now the applied heating power `qh_new = qh + dqh`,
read from `x[:, 1, 3:4]` (the `qh` component of the state at stage 1 after the first step).
Previously only the first element of the 1D control was extracted, which no longer corresponds
to the actual power after the reformulation.

### DIFF
--- a/leap_c/examples/hvac/acados_ocp.py
+++ b/leap_c/examples/hvac/acados_ocp.py
@@ -221,9 +221,10 @@ def export_parametric_ocp(
     qh_prev = ca.SX.sym("qh_prev")
     ocp.model.x = ca.vertcat(Ti, Th, Te, qh_prev)
 
-    # ── Input: qh [kW] ────────────────────────────────────────────────────────
+    # ── Input: [qh, t] where t = |qh| via LP reformulation ───────────────────
     qh = ca.SX.sym("qh")
-    ocp.model.u = qh
+    t = ca.SX.sym("t")
+    ocp.model.u = ca.vertcat(qh, t)
 
     # ── Discrete dynamics ─────────────────────────────────────────────────────
     Ad, Bd, Ed = transcribe_discrete_state_space(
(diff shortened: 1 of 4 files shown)


leap_c/examples/hvac/acados_ocp.py
```diff
@@ -214,16 +219,17 @@ def export_parametric_ocp(
     # Model
     ocp.model.name = name
 
-    # ── State: x_aug = [Ti, Th, Te, qh_prev] ─────────────────────────────────
+    # ── State: x_aug = [Ti, Th, Te, qh] ──────────────────────────────────────
```
JasperHoffmann, line 222: I think this way better :D

JasperHoffmann: Probably we should regularize it the RL loop. Still not sure, whether the integrator strategy was not too bad :)

PR #293, opened 2026-04-16
[`AcadosParameterManager` Refactor] Dropping `struct_symSX`

MazenAmria:
`AcadosParameterManager` currently uses CasADi's `struct_symSX` to build its parameter vectors. The struct requires all entries to be declared upfront in a single call, which prevents creating and returning CasADi symbols incrementally (i.e., one-by-one during parameter registration).

This ticket removes the dependency on the struct, and extends the parameter manager to allow adding parameters incrementally.

Changes summary:

1. Both `self.learnable_parameters` and `self.non_learnable_parameters` were replaced by a new internal class `_ParameterStore` that keeps track of: `symbols`, `defaults`, `sizes`, and optionally: `lb` and `ub`.
2. The store uses NumPy arrays for `defaults`, `lb` and `ub`.
3. `AcadosParameter` now has a default space that goes from `-np.inf` to `np.inf`, instead of having to check for missing space every time and replace it with a the default unbounded space.
4. The store allows for getting the symbols as a CasADI expression, and the flattened default values of the parameters.
5. Few properties were added to the parameter manager to directly get `p`, `p_global`, or the full parameter set, which is obtained by the property `p_full`. 
6. Some files other than `parameters.py` and `test_acados_parameters.py` have been changed, but that's because they have been relying on the internals of the parameter manager, which they shouldn't. The updated code delegates all the details into the parameter manager. Planners would only query `p`, `p_global`, or `p_full`, and the parameter manager handles the logic of gathering them. Same thing goes for the defaults.

### DIFF
--- a/leap_c/ocp/acados/parameters.py
+++ b/leap_c/ocp/acados/parameters.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass, field
-from typing import Any, Collection, Literal
+from typing import Any, Collection, Literal, Self
 from warnings import warn
 
 import casadi as ca
@@ -316,7 +316,7 @@ class AcadosParameterManager:
             if parameter.interface == "non-learnable":
                 self._store_non_learnable_parameter(parameter)
 
-    def add_parameter(self, parameter: AcadosParameter) -> "AcadosParameterManager":
+    def add_parameter(self, parameter: AcadosParameter) -> Self:
         """Adds a new parameter to the manager.
 
         This is a helper method for incrementally building the parameter manager, e.g. when
--- a/leap_c/ocp/acados/parameters.py
+++ b/leap_c/ocp/acados/parameters.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass, field
-from typing import Any, Collection, Iterable, Literal
+from typing import Any, Collection, Literal
 from warnings import warn
 
 import casadi as ca
@@ -7,7 +7,6 @@ import gymnasium as gym
 import numpy as np
 import torch
 from acados_template import AcadosOcp
-from casadi.tools import entry, struct, struct_symMX, struct_symSX
 
 
 @dataclass
@@ -44,6 +43,14 @@ class AcadosParameter:
 
     # Additional acados-specific field
     end_stages: list[int] = field(default_factory=list)
+
+    def __post_init__(self):
+        if self.space is None:
+            self.space = gym.spaces.Box(
+                low=-np.inf,
+                high=np.inf,
+                shape=np.shape(self.default),
+            )
 
 
 class AcadosParameterManager:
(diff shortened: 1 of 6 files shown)


MazenAmria:
UPDATE: the acceptance criteria of Task 0 includes incremental symbols creation, which hasn't been implemented within this PR.

leap_c/ocp/acados/parameters.py
```diff
@@ -174,106 +298,27 @@ def __init__(
...
+        # TODO: (Mazen) when parameters are added incrementally, we may add parameters that
+        # require an indicator after the construction of the parameter manager. We
+        # need to move this code into a point that is guaranteed to be run before the
+        # usage of the parameter manager for OCP solving. I guess that would be:
+        # `combine_non_learnable_parameter_values`?
+        if self._need_indicator:
+            indicator = AcadosParameter(
+                name="indicator", default=np.zeros(self.N_horizon + 1), interface="non-learnable"
+            )
+            self.add_non_learnable_parameter_entries(indicator)
```
MazenAmria: @JasperHoffmann @dirkpr The methods `add_learnable_parameter_entries` and `add_non_learnable_parameter_entries` can be used to add parameters incrementally. However, we will need to add the indicator if it wasn't required at the construction stage but then a new stagewise parameter required it. Two options to handle that:

1. on every parameter addition, we check if the indicator symbol exists or not, and add it if it's required.
2. after adding all the parameters, the caller of the parameter manager is required to call `combine_non_learnable_parameter_values` (not sure if that's how the API works, this is just my understanding of it), and then in that `combine_non_learnable_parameter_values` we create the indicators.

MazenAmria: Discussed with @JasperHoffmann and went with the 1st option.

dirkpr: Good. We need to have the sum of products of indicator variables (in p) and the corresponding p_global entries in the definition of the ocp. I think 1st option is the only one there.

leap_c/ocp/acados/parameters.py
```diff
@@ -305,20 +312,33 @@ def __init__(
...
-        # TODO: (Mazen) when parameters are added incrementally, we may add parameters that
-        # require an indicator after the construction of the parameter manager. We
-        # need to move this code into a point that is guaranteed to be run before the
-        # usage of the parameter manager for OCP solving. I guess that would be:
-        # `combine_non_learnable_parameter_values`?
-        if self._need_indicator:
-            indicator = AcadosParameter(
-                name="indicator", default=np.zeros(self.N_horizon + 1), interface="non-learnable"
+    def add_parameter(self, parameter: AcadosParameter) -> "AcadosParameterManager":
+        """Return a new AcadosParameterManager with the given parameter added.
```
dirkpr: Are we returning a new `AcadosParameterManager` or modifying the existing one?

leap_c/ocp/acados/parameters.py
```diff
@@ -305,20 +312,33 @@ def __init__(
...
+
+        This is a helper method for incrementally building the parameter manager, e.g. when
+        parameters are defined in different parts of the code.
+
+        Args:
+            parameter: The AcadosParameter to add.
+
+        Returns:
+            The same parameter manager, returned for chaining.
+        """
```
dirkpr, line 330: Is this function catching `end_stages` that are not compatible with the horizon? E.g. like here: https://github.com/MazenAmria/leap-c/blob/c86e8029f393cf4adee698030c2fa96959a0849f/leap_c/ocp/acados/parameters.py#L297-L305

dirkpr (changes requested):
Looks good! Only minor comments that should be quick to address.

dirkpr:
> 1. and in case of the learnable parameters: `lb` and `ub`

I remember that @JasperHoffmann earlier asked to remove lb, ub in favor of `gym.Space`. Did you guys discuss that?

dirkpr:
> Changes summary:

Also add some context to what the current status of the param-manager is and what the refactor contributes in improvements.

leap_c/ocp/acados/parameters.py
MazenAmria: We're modifying the existing one and returning it to allow method chaining:
```python
manager.add_parameter(p1).add_parameter(p2)
```

The `Return` field in the docstring is correct, while this description is wrong. I'll fix it.

MazenAmria:
> > 1. and in case of the learnable parameters: `lb` and `ub`
> 
> I remember that @JasperHoffmann earlier asked to remove lb, ub in favor of `gym.Space`. Did you guys discuss that?

@dirkpr We're planning to remove them, but the refactor plan has been divided into stages to avoid breaking all the tests at once. These will be removed in an upcoming PR.

MazenAmria:
@dirkpr updated the PR description, fixed the docstring, and added the missing validation.

dirkpr (approved):
Thanks, looks good!

leap_c/ocp/acados/parameters.py
FilippoAiraldi: Better to type hint the return as `typing.Self` then, no? Should be already available in Python 3.11

JasperHoffmann (approved):
Looks good! We can clean up when when we remove `learnable` and add outside parameter space management :)

PR #296, opened 2026-04-24
[`AcadosParameterManager` Refactor] Implementing `AcadosDiffOcp`

### LINKED ISSUE #294: Introducing `AcadosDiffOcp`
### Context

Currently, creating a parametric OCP requires a multi-step ceremony across several objects:

```python
# Current pattern (e.g., in CartPolePlanner.__init__):
params = create_cartpole_params(...)                              # 1. Create AcadosParameter list
param_manager = AcadosParameterManager(params, N_horizon)         # 2. Build manager
ocp = AcadosOcp()                                                 # 3. Create bare OCP
param_manager.assign_to_ocp(ocp)                                  # 4. Assign params to OCP
A = param_manager.get("A")                                        # 5. Get CasADi symbolic
ocp.model.disc_dyn_expr = A @ x + ...                             # 6. Use in formulation
```

Steps 1–5 are boilerplate that every example repeats. With the struct removed (#293), we can now offer a single `AcadosDiffOcp` class where `register_param` returns real CasADi symbols immediately — no separate `get()` call or explicit `finalize()` needed.

### Proposed interface

```python
class AcadosDiffOcp(AcadosOcp):
    """An AcadosOcp with built-in parameter registration for differentiable MPC.

    Requires N_horizon and optionally casadi_type at construction time so that register_param can
    eagerly build indicator expressions for split parameters.
    """

    def __init__(self, N_horizon: int, casadi_type: Literal["SX", "MX"] = "SX", *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.solver_options.N_horizon = N_horizon
        self.casadi_type = casadi_type
        self._param_manager: AcadosParameterManager = AcadosParameterManager(
            parameters=[],
            N_horizon=self.solver_options.N_horizon,
            casadi_type=self.casadi_type
        )
        self._finalized: bool = False

    def register_param(
        self,
        name: str,
        default: np.ndarray,
        space: gym.spaces.Space | None = None,
        differentiable: bool = False,
        end_stages: list[int] | None = None,
    ) -> ca.SX | ca.MX:
        """Register a parameter and return a CasADi symbolic for immediate use.

        The returned symbolic is a real CasADi SX (or MX) expression (not a placeholder).
        It can be used directly in cost, dynamics, and constraint expressions.

        Args:
            name: Parameter name.
            default: Default numerical value.
            space: Gymnasium space defining valid range (for learnable params).
            differentiable: If True, parameter supports sensitivities (learnable).
                If False, parameter is changeable at runtime but not differentiable
                (non-learnable).
            end_stages: Stage variation boundaries (see AcadosParameter.end_stages).

        Returns:
            A CasADi SX (or MX) symbolic expression for immediate use in OCP formulation.
        """
        if self._finalized:
            raise RuntimeError("Cannot register params after finalize().")

        interface = "learnable" if differentiable else "non-learnable"
        param = AcadosParameter(
            name=name, default=default, space=space,
            interface=interface, end_stages=end_stages or [],
        )
        self._param_manager.add_parameter(param)
        return self._param_manager.get(name)

    def finalize(self) -> None:
        """Build the AcadosParameterManager and assign params to the OCP.

        Called internally by AcadosDiffMpc — not by the user.
        Idempotent — safe to call multiple times.
        """
        self._param_manager.assign_to_ocp(self)
        self._finalized = True
```

### New OCP formulation pattern

```python
ocp = AcadosDiffOcp(N_horizon=96)
ocp.solver_options.tf = T_horizon

# register_param returns a real CasADi SX — use it immediately
q = ocp.register_param("q_diag_sqrt", default=np.ones(4), differentiable=True)
price = ocp.register_param("price", default=np.array([0.15]),
                           differentiable=False, end_stages=list(range(97)))
M = np.array([1.0])  # fixed params are just constants, no registration needed

# Formulate OCP directly — no finalize(), no get()
ocp.model.x = ca.SX.sym("x", 4)  # dirk: Are we running into problems when allowing ca.MX for parameters and now let the user use ca.SX? Possibly define CASADI_TYPE somewhere at start.
ocp.model.u = ca.SX.sym("u", 1)
ocp.model.cost_expr_ext_cost = q.T @ (ocp.model.x ** 2) + price * ocp.model.u
ocp.model.disc_dyn_expr = ...

# AcadosDiffMpc calls finalize() internally (just vertcat + assign)
diff_mpc = AcadosDiffMpcTorch(ocp, ...)
```

Compare with the old pattern:
```python
params = create_cartpole_params(...)
param_manager = AcadosParameterManager(params, N_horizon)
ocp = AcadosOcp()
param_manager.assign_to_ocp(ocp)
q = param_manager.get("q_diag_sqrt")
# ... formulate OCP ...
diff_mpc = AcadosDiffMpcTorch(ocp, ...)
# ... and separately pass param_manager around ...
```

### Where `param_manager` lives

The `param_manager` lives on `AcadosDiffOcp`. `AcadosDiffMpc` detects what it received:

```python
# In AcadosDiffMpcFunction.__init__:
def __init__(self, ocp, ...):
    if isinstance(ocp, AcadosDiffOcp):
        ocp.finalize()
        self.param_manager = ocp.param_manager
    else:
        # Old path — unchanged, param_manager passed separately or absent
        self.param_manager = None
```

Old code passing a plain `AcadosOcp` continues to work unchanged (no `param_manager` on OCP, old behavior preserved).

### Note on `recreate_dataclass`

With `register_param` returning symbolics directly, users construct dataclasses explicitly:

```python
gAw = ocp.register_param("gAw", default=..., differentiable=True)
Rea = ocp.register_param("Rea", default=..., differentiable=True)
params = HydronicDynamicsParameters(gAw=gAw, Rea=Rea, ...)
```

### Changes

1. **Create `AcadosDiffOcp`** in a new file `leap_c/ocp/acados/diff_ocp.py`.

2. **Implement `register_param` (eager, returns real symbols) and `finalize` (lightweight vertcat + assign).**

3. **Update `AcadosDiffMpcFunction.__init__`** to detect `AcadosDiffOcp`, call `finalize()`, and read `param_manager`. Old `AcadosOcp` path stays unchanged.

4. **Update `AcadosDiffMpcTorch.__init__`** similarly.

5. **Update one example** (e.g., `CartPolePlanner`) as a proof of concept. All other planners untouched.

### Files touched
- `leap_c/ocp/acados/diff_ocp.py` — new file
- `leap_c/ocp/acados/diff_mpc.py` — detect `AcadosDiffOcp`, read `param_manager`
- `leap_c/ocp/acados/torch.py` — detect `AcadosDiffOcp`, read `param_manager`
- `leap_c/examples/cartpole/acados_ocp.py` — proof of concept migration

### Acceptance criteria
- `register_param` returns real CasADi symbols that work in expressions immediately.
- `finalize()` correctly builds `p_global` / `p` via `ca.vertcat` without substitution.
- CartPole works end-to-end with the new `AcadosDiffOcp` pattern.
- All existing planners and tests pass unchanged (old `AcadosOcp` path preserved).

MazenAmria:
Currently, creating a parametric OCP requires a multi-step ceremony across several objects:

```python
# Current pattern (e.g., in CartPolePlanner.__init__):
params = create_cartpole_params(...)                              # 1. Create AcadosParameter list
param_manager = AcadosParameterManager(params, N_horizon)         # 2. Build manager
ocp = AcadosOcp()                                                 # 3. Create bare OCP
param_manager.assign_to_ocp(ocp)                                  # 4. Assign params to OCP
A = param_manager.get("A")                                        # 5. Get CasADi symbolic
ocp.model.disc_dyn_expr = A @ x + ...                             # 6. Use in formulation
```

Steps 1–5 are boilerplate that every example repeats. With the struct removed (#293), we can now offer a single `AcadosDiffOcp` class where `register_param` returns real CasADi symbols immediately — no separate `get()` call or explicit `finalize()` needed.

`acados_diff_ocp_tutorial.py` shows an example usage of the new interface. The `AcadosDiffOcp` contains the parameter manager and allows the `AcadosDiffMpc` to get the non-learnable parameters from the OCP instead of being passed by the planner.

This is probably something that needs to be discussed, but at this point there is no need for `AcadosPlanner`. A user can just create their own planner that uses `AcadosDiffMpc`. This makes the interface much cleaner, as the differentiable MPC is now treated as a standalone torch module, and the task-specific planner can include it just like any other module.

Closes https://github.com/leap-c/leap-c/issues/294

### DIFF
--- a/leap_c/ocp/acados/diff_mpc.py
+++ b/leap_c/ocp/acados/diff_mpc.py
@@ -15,6 +15,7 @@ from leap_c.ocp.acados.data import (
     collate_acados_flattened_batch_iterate_fn,
     collate_acados_ocp_solver_input,
 )
+from leap_c.ocp.acados.diff_ocp import AcadosDiffOcp
 from leap_c.ocp.acados.initializer import (
     AcadosDiffMpcInitializer,
     ZeroDiffMpcInitializer,
@@ -122,7 +123,7 @@ class AcadosDiffMpcFunction(DiffFunction):
 
     def __init__(
         self,
-        ocp: AcadosOcp,
+        ocp: AcadosOcp | AcadosDiffOcp,
         initializer: AcadosDiffMpcInitializer | None = None,
         sensitivity_ocp: AcadosOcp | None = None,
         discount_factor: float | None = None,
@@ -154,6 +155,8 @@ class AcadosDiffMpcFunction(DiffFunction):
 
         """
         self.ocp = ocp
+        if isinstance(ocp, AcadosDiffOcp):
+            ocp.finalize()
         self.forward_batch_solver, self.backward_batch_solver = (
             create_forward_backward_batch_solvers(
                 ocp=ocp,
@@ -213,6 +216,13 @@ class AcadosDiffMpcFunction(DiffFunction):
             - sol_value: The objective value solution, shape `(B, 1)`.
         """
         batch_size = x0.shape[0]
+
+        if isinstance(self.ocp, AcadosDiffOcp):
+            p_stagewise = (
+                self.ocp.parameter_manager.combine_non_learnable_parameter_values(batch_size)
+                if p_stagewise is None
+                else p_stagewise
+            )
 
         solver_input = AcadosOcpSolverInput(x0, u0, p_global, p_stagewise, p_stagewise_sparse_idx)
         ocp_iterate = None if ctx is None else ctx.iterate
--- a/leap_c/ocp/acados/diff_mpc.py
+++ b/leap_c/ocp/acados/diff_mpc.py
@@ -155,8 +155,6 @@ class AcadosDiffMpcFunction(DiffFunction):
 
         """
         self.ocp = ocp
-        if isinstance(ocp, AcadosDiffOcp):
-            ocp.finalize()
         self.forward_batch_solver, self.backward_batch_solver = (
             create_forward_backward_batch_solvers(
                 ocp=ocp,
--- a/leap_c/ocp/acados/diff_ocp.py
+++ b/leap_c/ocp/acados/diff_ocp.py
@@ -15,6 +15,9 @@ class AcadosDiffOcp(AcadosOcp):
     eagerly build indicator expressions for split parameters.
     """
 
+    # NOTE: We use a class-level dict to store parameter managers because `AcadosOcp` gets
+    # serialized into JSON, and `AcadosParameterManager` is not JSON-serializable. The dict
+    # is keyed by id(self) to allow multiple OCPs in the same process.
     _managers: dict[int, AcadosParameterManager] = {}
 
     def __init__(self, N_horizon: int, casadi_type: Literal["SX", "MX"] = "SX", **kwargs):
--- a/leap_c/ocp/acados/diff_ocp.py
+++ b/leap_c/ocp/acados/diff_ocp.py
@@ -1,0 +1,84 @@
+from typing import Literal
+
+import casadi as ca
+import gymnasium as gym
+import numpy as np
+from acados_template import AcadosOcp
+
+from leap_c.ocp.acados.parameters import AcadosParameter, AcadosParameterManager
+
+
+class AcadosDiffOcp(AcadosOcp):
+    """An AcadosOcp with built-in parameter registration for differentiable MPC.
+
+    Requires N_horizon and optionally casadi_type at construction time so that register_param can
+    eagerly build indicator expressions for split parameters.
+    """
+
+    _managers: dict[int, AcadosParameterManager] = {}
+
+    def __init__(self, N_horizon: int, casadi_type: Literal["SX", "MX"] = "SX", **kwargs):
+        super().__init__(**kwargs)
+        self.solver_options.N_horizon = N_horizon
+        self.casadi_type = casadi_type
+        type(self)._managers[id(self)] = AcadosParameterManager(
+            parameters=[], N_horizon=self.solver_options.N_horizon, casadi_type=self.casadi_type
+        )
+        self._finalized: bool = False
+
+    def register_param(
+        self,
+        name: str,
+        default: np.ndarray,
+        space: gym.spaces.Space | None = None,
+        differentiable: bool = False,
+        end_stages: list[int] | None = None,
+    ) -> ca.SX | ca.MX:
+        """Register a parameter and return a CasADi symbolic for immediate use.
+
+        The returned symbolic is a real CasADi SX (or MX) expression (not a placeholder).
+        It can be used directly in cost, dynamics, and constraint expressions.
+
+        Args:
+            name: Parameter name.
+            default: Default numerical value.
+            space: Gymnasium space defining valid range (for learnable params).
+            differentiable: If True, parameter supports sensitivities (learnable).
+                If False, parameter is changeable at runtime but not differentiable
+                (non-learnable).
+            end_stages: Stage variation boundaries (see AcadosParameter.end_stages).
+
+        Returns:
+            A CasADi SX (or MX) symbolic expression for immediate use in OCP formulation.
+        """
+        if self._finalized:
+            raise RuntimeError("Cannot register params after finalize().")
+
+        interface = "learnable" if differentiable else "non-learnable"
+        param = AcadosParameter(
+            name=name,
+            default=default,
+            space=space,
+            interface=interface,
+            end_stages=end_stages or [],
+        )
+        self.parameter_manager.add_parameter(param)
+        return self.parameter_manager.get(name)
+
+    def finalize(self) -> None:
+        """Build the AcadosParameterManager and assign params to the OCP.
+
+        Called internally by AcadosDiffMpc — not by the user.
+        Idempotent — safe to call multiple times.
+        """
+        self.parameter_manager.assign_to_ocp(self)
+        self._finalized = True
+
+    @property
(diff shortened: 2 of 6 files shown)


tutorial/acados_diff_ocp/acados_diff_ocp_tutorial.py
```diff
@@ -0,0 +1,87 @@
...
+
+    ocp.finalize()
+
+    ocp.constraints.x0 = np.array([20.0])
+
+    ocp.solver_options.tf = N_HORIZON * dt
+    ocp.solver_options.N_horizon = N_HORIZON
+    ocp.solver_options.qp_solver = "PARTIAL_CONDENSING_HPIPM"
+    ocp.solver_options.hessian_approx = "EXACT"
+    ocp.solver_options.integrator_type = "DISCRETE"
```
MazenAmria, line 69: In the future, such configurations need to be encapsulated. The `AcadosDiffOcp` should be an interface that allows configuring everything related to a differentiable OCP, and it uses a parameter manager under the hood to enable the features that acados does not natively support.

leap_c/ocp/acados/diff_ocp.py
```diff
@@ -0,0 +1,87 @@
...
+    """An AcadosOcp with built-in parameter registration for differentiable MPC.
+
+    Requires N_horizon and optionally casadi_type at construction time so that register_param can
+    eagerly build indicator expressions for split parameters.
+    """
+
+    # NOTE: We use a class-level dict to store parameter managers because `AcadosOcp` gets
+    # serialized into JSON, and `AcadosParameterManager` is not JSON-serializable. The dict
+    # is keyed by id(self) to allow multiple OCPs in the same process.
+    _managers: dict[int, AcadosParameterManager] = {}
```
JasperHoffmann: wow, I guess it works 😄

tutorial/acados_diff_ocp/acados_diff_ocp_tutorial.py
JasperHoffmann: Agreed :) We should open a new issue.

leap_c/ocp/acados/diff_mpc.py
```diff
@@ -122,7 +123,7 @@ class AcadosDiffMpcFunction(DiffFunction):
 
     def __init__(
         self,
-        ocp: AcadosOcp,
+        ocp: AcadosOcp | AcadosDiffOcp,
         initializer: AcadosDiffMpcInitializer | None = None,
         sensitivity_ocp: AcadosOcp | None = None,
```
JasperHoffmann, line 128: At one point we should maybe remove the sensitivity OCP and also require `AcadosDiffOcp`. Will collect as an Issue :)

tutorial/acados_diff_ocp/acados_diff_ocp_tutorial.py
JasperHoffmann: Just an idea, but we probably can lazy construct `p_global` and `p`, and in the long term do not need a finalize ^^ Maybe we can overwrite the getter method and then internally call finalize? The setter method we can then block away, because you should not change it. Does this make sense to you?

JasperHoffmann (approved):
The new interface looks great 😍 Just small comment on finalize. As far as I know we just need to build `p` and `p_global` so probably we can just write custom getter and setter and block some stuff.

MazenAmria:
> need to build `p` and `p_global` so probably we can just write custom getter and setter and block some stuff.

The parameter manager already contains some properties for `p`, `p_global`, and the flattened default values of them (which is all what gets assigned to the OCP in the finalize step). We can probably simplify things a bit.

leap_c/ocp/acados/diff_ocp.py
```diff
@@ -68,20 +97,29 @@ def register_param(
...
-
     @property
     def parameter_manager(self) -> AcadosParameterManager:
         """Return the AcadosParameterManager for this OCP."""
-        return type(self)._managers[id(self)]
+        return type(self)._managers[self.manager_id]
 
-    def __del__(self):
-        """Clean up the parameter manager for this OCP."""
-        type(self)._managers.pop(id(self), None)
```
MazenAmria: It seems like the solver do not use the same OCP instance but instead reinstantiate it from the serialized JSON. Hence, we can't clean up the managers, but we keep them in a class-level dict while the UUID key is serialized to JSON with the OCP, so that it can be referenced again when it's reinstantiated.

FilippoAiraldi:
Some very minor comments on `AcadosDiffOcp`.

- there is no need to `type(self)._managers`, we can just access the field as `self._managers`
- what about, instead of `uuid64`, we use `id(self)`? Should be slightly cheaper to build than an uuid. It is guaranteed to return a unique integer, though two parameter managers with non-overlapping lifetimes may have the same ID. To solve the issue we could just have `weakref.finalize(self, self._managers.pop, self.manager_id, None)` to make sure the id is popped out of the dict when a manager is GC'd. The code would then look something like
```python
def __init__(...) -> None:
    ...
    self.manager_id = id(self)
    self._managers[self.manager_id] = AcadosParameterManager(...)
    weakref.finalize(self, self._managers.pop, self.manager_id, None)

...

@property
def parameter_manager(self) -> AcadosParameterManager:
    return self._managers[self.manager_id]
```

but maybe this is overengineering.

MazenAmria:
@FilippoAiraldi I was leaning towards `weakref` initially, being the correct solution to this problem, but I thought it may add another level of complexity for people reading the code. Now I'm convinced to use it again.

PR #298, opened 2026-05-06
Allow multiple steps in ERK4 integration

josipkh:
Extends the integrator utility to avoid loss of accuracy if the required number of steps is larger than 1.

Also aligns the citation in the README with the .bib file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,30 +24,12 @@ Open a new thread or browse the existing ones on the [GitHub discussions](https:
 ## Citing
 If you are using code from this repository in your work, please use the following citation for now
 ```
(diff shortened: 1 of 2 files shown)


dirkpr:
Thanks! We have an implementation of an ERK in a draft PR that should see a merge soon and which I think is somewhat more readable: https://github.com/leap-c/leap-c/pull/295/changes#diff-5e43083742bcab84d541b482b5c88458d6803a5ea3288c676e52eefb753c818e .

josipkh:
Right, didn't catch that...
Thanks for the quick feedback

PR #299, opened 2026-05-07
Allow QP scaling in the forward solver

josipkh:
Allows the forward solver to use QP scaling, which [should be disabled](https://github.com/acados/acados/blob/4b4bb36028b0a3ead1d1513bc310e5b5f2f7f54a/interfaces/acados_template/acados_template/acados_ocp_options.py#L2346) for correct sensitivity computation.
Also aligns the citation entry.


JasperHoffmann (approved):
Thank you :)

PR #301, opened 2026-05-08
[`AcadosParameterManager` Refactor] Using `splits` instead of `end_stages` in `AcadosParameter`

MazenAmria:
* `AcadosParameter` now uses `splits` instead of `end_stages`, which supports different options:
  * `list[int]` matches the original behavior of `end_stages`.
  * `int` defines almost-equal-sized splits, when `N_horizon + 1` is not divisible by the provided number of splits, then we expand the first `remainder` stages by 1. The number of splits may not exceed the number of stages in the parameter manager (those being `N_horizon + 1`).
  * `"stagewise"` equivalent to using `list(range(N_horizon + 1))`.
  * `"global"` parameter is not stage-varying.
* `end_stages` is still kept for backward compatibility, old tests are still passing.
* A `DeprectationWarning` was added to discourage users from using `end_stages`.
* Added unit tests for the splits validation and the start/end calculations.

### DIFF
--- a/leap_c/ocp/acados/diff_ocp.py
+++ b/leap_c/ocp/acados/diff_ocp.py
@@ -68,6 +68,7 @@ class AcadosDiffOcp(AcadosOcp):
         default: np.ndarray,
         space: gym.spaces.Space | None = None,
         differentiable: bool = False,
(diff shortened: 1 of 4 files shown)


JasperHoffmann (approved):
Looks good!

PR #302, opened 2026-05-21
[`AcadosDiffMpcTorch` Refactor] Using dict-based parameters and deprecating the old interface.

MazenAmria:
* The old interface used to pass `p_global`, `p_stagewise` and `p_stagewise_sparse_idx`. This has been deprecated in favor of a cleaner parameter dictionary.
* The dictionary gets split and flattened in `AcadosDiffMpcTorch` and then sent to the `AcadosDiffMpcFunction` as normal.
* `AcadosDiffMpcTorch` requires the parameter manager to be able to split the dictionary. Hence, we deprecated the construction of `AcadosDiffMpcTorch` with an `AcadosOcp` and encouraged using `AcadosDiffOcp` which comes with a parameter manager.
* The user passes a dictionary of parameters with clear semantics, instead of flattened and concatenated vectors.


leap_c/ocp/acados/planner.py
```diff
@@ -60,7 +60,7 @@ def forward(
         parameters are automatically obtained from the `param_manager`.
         """
         p_stagewise = self.param_manager.combine_non_learnable_parameter_values(obs.shape[0])
```
MazenAmria: This need to be encapsulated within the differentiable MPC. Exposing it to the planner defeats the purpose of passing parameters as a dictionary.

I would like to hear your opinions on that.

MazenAmria: @JasperHoffmann @dirkpr

JasperHoffmann: Sounds good! The long term plan is to get rid of `AcadosPlanner` to reduce complexity ;)

leap_c/ocp/acados/parameters.py
```diff
@@ -236,6 +236,16 @@ def learnable_default_flat(self) -> np.ndarray:
     def non_learnable_default_flat(self) -> np.ndarray:
         return self._non_learnable_parameter_store.get_values()
 
+    @property
```
FilippoAiraldi, line 212: Should we consider caching these properties via `functools.cached_property` (to avoid computing these list comprehensions at every `AcadosDiffMpcTorch.forward` call, which can get pretty hot)?

The question is whether the `AcadosParameterManager` should be considered frozen once built, or whether we want it to remain modifiable. 
- If the answer is the former, then there is no downside to caching. De facto, we could achieve the same by computing these heavier properties only once in `AcadosParameterManager.__init__` and subsequently retrieving their values as attributes.
- If the answer is the latter, caching can still be achieved but we need to put in a place a caching-invalidation mechanism (can be done, I did it in another project, but can be annoying).

leap_c/ocp/acados/torch.py
```diff
@@ -66,6 +67,10 @@ def __init__(
 
         """
         super().__init__()
+        if not isinstance(ocp, AcadosDiffOcp):
+            warnings.warn(
```
FilippoAiraldi: Better to use class names and be more specific about the type of warning raised

```python
if not isinstance(ocp, AcadosDiffOcp):
    warnings.warn(
        f"Passing an `{AcadosOcp.__name__}` instance is deprecated. Please use  `{AcadosDiffOcp.__name__}`.",
        DeprecationWarning,
        stacklevel=2,
    )
```

We could also add two `@overload`ed definitions of `__init__` (one with `AcadosOcp`, the other with `AcadosDiffOcp`) and decorate the former with `@typing_extensions.deprecated` (in case `typing_extensions` is already a dependency of ours).

tests/leap_c/ocp/acados/test_acados_torch.py
```diff
@@ -804,3 +804,80 @@ def test_backward_nmpc(
     check_gradients(diff_mpc_indefinite_hess, n_batch, max_batch_size, dtype, noise_scale)
     print("Check stagewise diff_mpc===============================")
     check_gradients(diff_mpc_indefinite_hess_stagewise, n_batch, max_batch_size, dtype, noise_scale)
+
+
+def create_simple_diff_mpc(N_horizon: int = 5):
+    """Creates a minimal differentiable MPC for testing parameter gradients."""
+    import casadi as ca
```
FilippoAiraldi: why are there local imports in `create_simple_diff_mpc` and `test_params_dict_backward_simple`? Also, some of the imports are repeated (e.g., `torch`, `np`, `AcadosDiffMpcTorch`).

leap_c/ocp/acados/torch.py
```diff
@@ -120,6 +136,42 @@ def forward(
...
+                "When providing `params`, it must contain all parameters defined in the ocp object."
+                f" Expected global parameters: {global_params},"
+                f" stagewise parameters: {stagewise_params}."
+            )
+            p_global = [params[name] for name in global_params if name in params]
+            p_global = torch.cat(p_global, dim=-1) if p_global else None
+            p_stagewise = [params[name] for name in stagewise_params if name in params]
+            p_stagewise = torch.cat(p_stagewise, dim=-1) if p_stagewise else None
+            p_stagewise_sparse_idx = [
+                params[name + "_sparse_idx"]
```
FilippoAiraldi: Why not add the list of keys from `p_stagewise_sparse_idx` as a property to `AcadosParameterManager`? E.g.,

```python
class AcadosParameterManager:
    ...

    @property
    def stagewise_sparse_idx_names(self) -> list[str]:
        return [
            name + "_sparse_idx" for name, param in self.parameters.items() 
            if param.interface == "non-learnable"
        ]
```

Or at least avoid the double string concatenation with a walrus 🦭

```python
p_stagewise_sparse_idx = [
    params[key]
    for name in stagewise_params
    if (key := name + "_sparse_idx") in params
]
```

leap_c/ocp/acados/torch.py
```diff
@@ -120,6 +136,42 @@ def forward(
...
+            global_params = self.parameter_manager.global_parameter_names
+            stagewise_params = self.parameter_manager.stagewise_parameter_names
+            assert set(global_params) <= set(params.keys()) and set(stagewise_params) <= set(
+                params.keys()
+            ), (
+                "When providing `params`, it must contain all parameters defined in the ocp object."
+                f" Expected global parameters: {global_params},"
+                f" stagewise parameters: {stagewise_params}."
+            )
+            p_global = [params[name] for name in global_params if name in params]
```
FilippoAiraldi: No need for `if name in params`, both on line R162 and R164; the assert already guarantees that these names are in `params`.

FilippoAiraldi (dismissed):
Minor tweaks to code, and a question on whether to cache or not.

leap_c/examples/chain/planner.py
```diff
@@ -79,34 +76,23 @@ def __init__(
...
-        param_manager = AcadosParameterManager(
-            parameters=params,
-            N_horizon=self.cfg.N_horizon,  # type:ignore
-        )
 
         fix_point = np.zeros(3)
 
         # find resting reference position
-        rest_length = param_manager.parameters["L"].default[0]
+        rest_length = 0.033
```
JasperHoffmann, line 86: Maybe make it a global constant in the module?

JasperHoffmann: This was unfortunately always a little bit annoying.

leap_c/examples/chain/acados_ocp.py
```diff
@@ -101,17 +75,18 @@ def export_parametric_ocp(
...
+
+    # Physical parameters (sunsetted "fix" interface)
     dyn_param_dict = OrderedDict(
         [
-            ("D", param_manager.get("D")),
-            ("L", param_manager.get("L")),
-            ("C", param_manager.get("C")),
-            ("m", param_manager.get("m")),
-            ("w", param_manager.get("w")),
+            ("D", np.repeat([1.0, 1.0, 1.0], n_mass - 1)),
```
JasperHoffmann, line 82: Same as in chain, maybe collect it in dataclass :)?

leap_c/examples/hvac/acados_ocp.py
```diff
@@ -55,201 +139,69 @@ def make_default_hvac_params(
...
+    # Input: qh [kW]
     qh = ca.SX.sym("qh")
     ocp.model.u = qh
 
-    # ── Discrete dynamics ─────────────────────────────────────────────────────
+    # Discrete dynamics
     Ad, Bd, Ed = transcribe_discrete_state_space(
-        dt=dt,
-        params=param_manager.recreate_dataclass(HydronicDynamicsParameters),
+        dt=900.0,
```
JasperHoffmann, line 179: dataclass?

leap_c/ocp/acados/planner.py
```diff
@@ -3,14 +3,14 @@
...
 import torch
-from numpy import ndarray
 
 from leap_c.controller import CtxType
 from leap_c.ocp.acados.diff_mpc import AcadosDiffMpcCtx, collate_acados_diff_mpc_ctx
 from leap_c.ocp.acados.parameters import AcadosParameterManager
 from leap_c.ocp.acados.torch import AcadosDiffMpcSensitivityOptions, AcadosDiffMpcTorch
 from leap_c.planner import ParameterizedPlanner, SensitivityOptions
 
+# TODO Jasper: Needs to be updated if we go to a dictionary based sensitivity derivation.
```
JasperHoffmann, line 13: We should probably update this :)

JasperHoffmann: This should be simplified we should not offer this anymore in the abstract interface for now?

leap_c/examples/chain/acados_ocp.py
```diff
@@ -20,71 +21,44 @@
...
     n_mass: int = 5,
-) -> AcadosOcp:
+) -> tuple[AcadosOcp, AcadosParameterManager]:
     ocp = AcadosOcp()
 
     ocp.solver_options.N_horizon = N_horizon
     ocp.solver_options.tf = T_horizon
 
-    param_manager.assign_to_ocp(ocp)
+    manager = AcadosParameterManagerTorch(N_horizon=N_horizon)
```
JasperHoffmann, line 38: The type is a little bit weird, currently we need to import both. Maybe we should simplify this?

leap_c/ocp/acados/parameters.py
```diff
@@ -175,39 +165,22 @@ def get_symbols(self) -> ca.SX | ca.MX:
...
 
 
-class AcadosParameterManager:
-    """Manager class for handling acados parameters according to their specifications.
-
-    In particular, this handles stage-varying learnable parameters, which is not available
-    out-of-the-box in acados.
-
-    Basic usage inside the AcadosOcp formulation:
+class AcadosParameterManager(ABC):
```
JasperHoffmann, line 168: I think we should make no subclasses for now and just add the specific combine_learnable_params in to this class :) The imports can be done in the method itself, if we want to allow to have partial installation requiring not both jax and torch!

leap_c/ocp/acados/parameters.py
JasperHoffmann: I think we can assume that this becomes frozen :) However, I still could see that this can introduce nasty bugs.  I think I would even mark them for deletion. As default values should be managed outside of this class!

leap_c/ocp/acados/parameters.py
```diff
@@ -291,90 +270,63 @@ def _store_non_learnable_parameter(self, parameter: AcadosParameter) -> None:
...
-                self._store_non_learnable_parameter(parameter)
+        self._finalized = False
 
-    def add_parameter(self, parameter: AcadosParameter) -> Self:
-        """Adds a new parameter to the manager.
+    def register_parameter(
+        self,
+        name: str,
+        default: np.ndarray,
+        space: gym.spaces.Space | None = None,
```
JasperHoffmann, line 297: This should be removed in a future PR.

JasperHoffmann (approved):
There are still some changes we should do in the future :)

leap_c/ocp/acados/parameters.py
dirkpr: note: will also affect AcadosDiffMpcTorch.param_space

leap_c/ocp/acados/parameters.py
```diff
@@ -175,39 +165,22 @@ def get_symbols(self) -> ca.SX | ca.MX:
...
     **Stage-varying learnable parameters** (``splits`` are not ``"global"``) are implemented via a
     one-hot *indicator* vector that is appended to the non-learnable parameters.  At stage ``k``
-    only ``indicator[k]`` is 1; ``get()`` returns a weighted sum over all stage blocks so the
+    only ``indicator[k]`` is 1; :meth:`get` returns a weighted sum over all stage blocks so the
     same symbolic expression evaluates to the correct block value at every stage.
 
     .. warning::
         If you forget to set the indicator correctly in
-        ``combine_non_learnable_parameter_values()``, every stage will silently evaluate to zero
+        :meth:`combine_non_learnable_parameters`, every stage will silently evaluate to zero
```
dirkpr, line 183: The indicator-part of the parameters should be correct by default, right?

dirkpr (approved):

FilippoAiraldi:
Noo, I was doing my review 🥲

Found a couple of issues. Will open new issues or make PRs myself

JasperHoffmann:
> Noo, I was doing my review 🥲
> 
> Found a couple of issues. Will open new issues or make PRs myself

Feel free to do so :) There are still a lot of improvement points.

PR #303, opened 2026-05-26
[CI Fixes] Building acados and fixing dependencies

MazenAmria:
* Fetching artifacts of acados runs causes the pipeline to fail when the runs' artifacts are removed due to github actions' retention policy. We build acados in our pipeline instead. This also ensures consistency between documented installation process and tested source code. Our tests now run on an acados version built with the exact same options documented in our installation guide.
* `requests-cache` was failing with `name 'RequestsCookieJar' is not defined`. Github Actions was installing a newer version of `requests` causing HVAC tests to fail in the pipeline. Verified locally by fixing `requests` version, removing all the cached responses, and bisecting HVAC tests to identify the failing version `>=2.34.0`.


.github/workflows/documentation.yml
FilippoAiraldi, line 1: For the moment, would it perhaps be convenient to refactor `prepare_acados` out of `tests.yml` and `documentation.yml` into a standalone reusable workflow that the other two workflows can call? See https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows

.github/actions/build-acados/action.yml
```diff
@@ -0,0 +1,31 @@
...
+
+    - name: Compile acados Manually
+      if: steps.cache-acados.outputs.cache-hit != 'true'
+      working-directory: ${{ github.workspace }}/leap_c/external/acados
+      shell: bash
+      run: |
+        mkdir -p build
+        cd build
+        cmake -DACADOS_WITH_OPENMP=ON -DACADOS_NUM_THREADS=1 ..
+        make install -j4
```
FilippoAiraldi: Maybe `-j"$(nproc)"` instead of hard-coding 4 parallel jobs with `-j4`?

Also, if we want to use a more modern CMake syntax, we could replace these four lines with
```bash
cmake -S . -B build -DACADOS_WITH_OPENMP=ON -DACADOS_NUM_THREADS=1
cmake --build build --target install -- -j"$(nproc)"
```

FilippoAiraldi (commented):
I have just two suggestions for refactoring.

On a more general note, do we really find it necessary to cache the compiled `acados` binaries, instead of building every time? On a repository of mine where I build `acados` for every test (no caching), the build time clocks at around 1min and 10/20sec. In comparison, `leap-c`'s tests run in 7/8min.

MazenAmria:
> I have just two suggestions for refactoring.
> 
> On a more general note, do we really find it necessary to cache the compiled `acados` binaries, instead of building every time? On a repository of mine where I build `acados` for every test (no caching), the build time clocks at around 1min and 10/20sec. In comparison, `leap-c`'s tests run in 7/8min.

@FilippoAiraldi I wouldn't say necessary, but I'd always be in favor of optimizing the pipeline time.

FilippoAiraldi:
@MazenAmria I was just worrying about over-engineering, after reading your comment. But for now the caching does not seem to add much complexity :)

FilippoAiraldi (changes requested):
Just two other small changes

1) unless I am misunderstanding, shouldn't the vars `ACADOS_SOURCE_DIR`, `ACADOS_INSTALL_DIR`, `LD_LIBRARY_PATH` be exported in either `action/build-acados` or `workflows/prepare_acados.yml`, instead of being exported in `actions/setup-leap-c`? As far as I can tell, these variables are needed for the installation of `acados` rather than `leap-c` (at least, not directly).
2) with 
```bash
echo "LD_LIBRARY_PATH=${{github.workspace}}/leap_c/external/acados/lib" >> $GITHUB_ENV
```
we are overwriting any pre-existing `LD_LIBRARY_PATH` with a new value, whereas we should be appending to it the new path
```bash
echo "LD_LIBRARY_PATH=${{github.workspace}}/leap_c/external/acados/lib:$LD_LIBRARY_PATH" >> $GITHUB_ENV
```

Everything else seems in order to me; don't know if @dirkpr or @JasperHoffmann want to chip in.

MazenAmria:
> Just two other small changes

1. The documentation isn't explicit about it, but it states that Reusable Workflows allow "Specifying runners" (https://docs.github.com/en/actions/concepts/workflows-and-actions/reusing-workflow-configurations#reusable-workflows-versus-composite-actions), so I'd assume they're running on a different runner than the downstream jobs (tests and docs), and those jobs would not detect the exported environment variables. The current approach was to first ensure that acados is cached with a single job, and then trigger all the testing jobs, and those jobs would again restore the cache, so that concurrent testing jobs do not duplicate the effort of building acados when all of them face a cache miss. Honestly, caching is justified but that bit of avoiding duplicate builds on the first cache miss is over-engineered. I'll remove the reusable workflow and just rely on a composite action to restore the cached acados build, build on cache miss, install Tera, then export the environment variables.
2. Good Catch!

.github/workflows/documentation.yml
JasperHoffmann: That sounds nice :)

FilippoAiraldi (approved):
I like it :) The execution flow is easy to follow. 

Just a note, all the logs from the latest action for this PR report `Cache not found for input keys: Linux-acados-6111f01f2220a241057b4d56d8fe3a2cddd0309f`. Did we expect all cache misses until merged?

MazenAmria:
> Did we expect all cache misses until merged?

No, the caches are scoped across different commits inside the same PR, as they're listed under the same ref: `refs/pull/<pr_number>/merge`. This cache miss is caused by changing the [cache version](https://github.com/actions/cache#cache-version) due to changing the paths (I have appended the `/bin` path in the last commit).

Both versions are cached and appear in the cache list API:

```json
{
  "total_count": 5,
  "actions_caches": [
    {
      "id": 4620834724,
      "ref": "refs/pull/303/merge",
      "key": "Linux-acados-6111f01f2220a241057b4d56d8fe3a2cddd0309f",
      "version": "35e0e78ca3a763cf5d69ada54ac78ca4ab02c3e4f27d513d56017b9629eb0cf9",
      "last_accessed_at": "2026-05-27T23:10:54.955941000Z",
      "created_at": "2026-05-27T23:10:54.955941000Z",
      "size_in_bytes": 222
    },
    // ...
    {
      "id": 4592716090,
      "ref": "refs/pull/303/merge",
      "key": "Linux-acados-6111f01f2220a241057b4d56d8fe3a2cddd0309f",
      "version": "4f27f610d36813c7387baf85e5b577253b58607e95bfa8cbdcb6232028d44c91",
      "last_accessed_at": "2026-05-26T20:55:46.562328000Z",
      "created_at": "2026-05-26T20:55:06.549007000Z",
      "size_in_bytes": 5156346
    }
  ]
}
```

[This action](https://github.com/leap-c/leap-c/actions/runs/26473154660/job/77951710941) for example encounters a cache hit.

PR #315, opened 2026-06-28
Build parameter space outside of parameter manager

### LINKED ISSUE #307: Move gym space management outside the AcadosParameterManager
This should be removed in a future PR.

_Originally posted by @JasperHoffmann in https://github.com/leap-c/leap-c/pull/302#discussion_r3481398595_

dirkpr:
# Handle parameter space outside `AcadosParameterManager`

Pulls parameter-space/bounds handling out of `AcadosParameterManager` and moves it to `AcadosDiffMpcTorch`. The manager now owns only parameter *values*; the *space* (bounds) is built externally as a `gym.spaces.Dict` keyed by learnable parameter name and threaded into the controller.

The manager mixed two concerns — parameter values (its job) and space/bounds (used only by the RL layer). Separating them aligns space handling with the dict-based parameter API. Fix #307.

## Changes
- **`AcadosParameterManager`**: removed `get_param_space()`; `register_parameter`'s `space=` argument is no longer used to build the space.
- **`AcadosDiffMpcTorch`**: new required `param_space: gym.spaces.Dict` argument; the `param_space` property returns it verbatim. A later PR will remove `param_space` from **`AcadosDiffMpcTorch`**.
- **New helpers**:
  - `stage_expanded_box(...)` in `parameters.py` — tiles single-stage bounds across stage blocks for stage-varying params (no-op for `"global"`).
  - `flatten_param_space(...)` in `utils/gym.py` — flattens a `Dict` (or passes a `Box` through) into the canonical flat learnable vector.
- **Examples** (cartpole, chain, mass_spring_damper, pointmass): each `acados_ocp.py` builds the space as a list of `(name, Box)` tuples in registration order and returns a `gym.spaces.Dict`; planners thread it through.
- **Consumers** flatten the Dict where a flat space is needed: `mpc_actor`, `sac_zop`/`sac_fop`, `crossq_zop`, `run_smac_tuning`.


leap_c/examples/cartpole/acados_ocp.py
```diff
@@ -37,19 +37,32 @@ def export_parametric_ocp(
 
     manager = AcadosParameterManager(N_horizon=N_horizon)
 
+    spaces: list[tuple[str, gym.spaces.Box]] = []
```
JasperHoffmann, line 40: Works. Just a short question should we always enforce DictSpaces, we could haven make it simpler and say you can also have a box if you only have a single Param. Then `ParameterizedPlanner` does not need to enforce DictSpaces, but it can be also flat. From the `mpc_actor` side we can handle flexibly handle both? What do you think?

leap_c/ocp/acados/torch.py
```diff
@@ -83,13 +89,14 @@ def __init__(
             verbose=verbose,
         )
         self.parameter_manager = parameter_manager
+        self._param_space = parameter_space
         self.autograd_fun = create_autograd_function(self.diff_mpc_fun)
         self.dtype = torch.get_default_dtype() if dtype is None else dtype
 
     @property
-    def param_space(self) -> gym.Space:
+    def param_space(self) -> gym.spaces.Dict:
```
JasperHoffmann, line 97: Same discussion, we can make it even more flexible :) Even though if we enforce names it can be good for logging.

JasperHoffmann (approved):
Looks good. Only added some discussion points.

leap_c/examples/cartpole/acados_ocp.py
dirkpr: I see the point that enforcing a `spaces.Dict` can be in the way of a single parameter vector. A quick workaround there could be to just have one key, e.g. `param`, with one box?

JasperHoffmann: yes I think we can actually do both and maybe it is nice to have a flat version at least in one of the examples :)

PR #316, opened 2026-06-28
Refactor param management in examples

JasperHoffmann:
This PR makes parameter declaration local, explicit, and decoupled for each example directly in the definition of the ocp.

## Changes

-  `param_space` is built explicitly per example. `AcadosDiffMpcTorch` is a plain `torch.nn.Module` (it no longer owns param_space), and `AcadosPlanner` is removed. Each `def export_parametric_ocp` constructs the `gym.spaces.Space` itself and returns `(ocp, manager, param_space, default_param)`.
- Splits are set inline at registration not globally per-controller param_interface.
- Generic `flatten/unflatten` replace `_param_to_dict`.


leap_c/examples/cartpole/planner.py
```diff
@@ -98,11 +103,40 @@ def __init__(
...
+        return self.diff_mpc(x0=obs, u0=action, params=params, ctx=ctx)
+
+    def sensitivity(self, ctx: AcadosDiffMpcCtx, name: SensitivityOptions) -> np.ndarray:
+        return acados_sensitivity(self.diff_mpc, ctx, name)
+
+    @property
+    def param_space(self):
+        return self._param_space
+
+    def default_param(self, obs: np.ndarray | torch.Tensor | None = None) -> np.ndarray:
```
JasperHoffmann, line 124: This can be even simplified I guess by setting None but not sure whether we want to use it somewhere for logging :D?

leap_c/examples/cartpole/planner.py
```diff
@@ -27,7 +30,7 @@ class CartPolePlannerConfig:
         x_threshold: Bounds of the box constraints of the maximum absolute position
             of the cart [m] (soft/slacked constraint)
         cost_type: The type of cost to use, either "EXTERNAL" or "NONLINEAR_LS".
-        param_interface: Determines the exposed parameter interface of the planner.
+        param_splits: Determines the exposed parameter interface of the planner.
```
dirkpr: `param_splits` needs updated doc?

leap_c/examples/cartpole/planner.py
```diff
@@ -85,8 +89,9 @@ def __init__(
         """
         self.cfg = CartPolePlannerConfig() if cfg is None else cfg
 
-        ocp, param_manager, param_space = export_parametric_ocp(
-            param_interface=self.cfg.param_interface,
+        super().__init__()
+        ocp, param_manager, param_space, default_param = export_parametric_ocp(
+            param_splits=self.cfg.param_splits,
```
dirkpr: Is there a reason we define `param_splits` in the `CartPolePlannerConfig`? I think we should just define it for each param explicitly when using `manager.register_param`. The function was introduced for that reason to avoid the multi ceremony.

leap_c/examples/cartpole/acados_ocp.py
```diff
@@ -21,14 +17,14 @@
...
-    param_interface: CartPoleAcadosParamInterface,
+    param_splits: ParamSplits,
     cost_type: CartPoleAcadosCostType = "NONLINEAR_LS",
     name: str = "cartpole",
     Fmax: float = 80.0,
     x_threshold: float = 2.4,
     N_horizon: int = 50,
     T_horizon: float = 2.0,
-) -> tuple[AcadosOcp, AcadosParameterManager, gym.spaces.Dict]:
+) -> tuple[AcadosOcp, AcadosParameterManager, gym.spaces.Box, np.ndarray]:
```
dirkpr, line 27: Why do we return `default_param` explicitly? They are registered with the `param_manager`.

leap_c/examples/cartpole/acados_ocp.py
```diff
@@ -37,30 +33,20 @@ def export_parametric_ocp(
...
-                gym.spaces.Box(
-                    low=np.array([-2.0 * np.pi]),
-                    high=np.array([2.0 * np.pi]),
-                    dtype=np.float64,
-                ),
-                xref1_splits,
-                N_horizon,
-            ),
-        )
+    default_param = manager.default_param_dict(["xref1"])["xref1"]
```
dirkpr: `default_param` seems misleading here. `default_differentiable_param`?

leap_c/examples/cartpole/planner.py
```diff
@@ -45,15 +48,15 @@ class CartPolePlannerConfig:
     x_threshold: float = 2.4
 
     cost_type: CartPoleAcadosCostType = "NONLINEAR_LS"
-    param_interface: CartPoleAcadosParamInterface = "global"
+    param_splits: ParamSplits = "global"
```
dirkpr: I think we should remove `param_splits` from the config and use it explicitly at `register_param`, see also next comment.

leap_c/examples/cartpole/planner.py
```diff
@@ -98,11 +103,40 @@ def __init__(
...
             discount_factor=self.cfg.discount_factor,
             export_directory=export_directory,
             n_batch_init=self.cfg.n_batch_init,
             num_threads_batch_solver=self.cfg.num_threads_batch_solver,
             dtype=self.cfg.dtype,
         )
-        super().__init__(param_manager=param_manager, diff_mpc=diff_mpc)
+        self.param_manager = param_manager
+        self.diff_mpc = diff_mpc
+        self._param_space = param_space
```
dirkpr, line 110: Do we want to keep the planner? If not, I would see  `_param_space` in `AcadosDiffMpcTorch`

leap_c/examples/cartpole/planner.py
dirkpr: Not sure if we are clean on `default_param`. Thinking of other parameters that are not differentiable. At least we should make it clear in the documentation that at this stage, we only consider differentiable params. And other parameters need to be handled explicitly in the planner.

leap_c/examples/chain/planner.py
```diff
@@ -42,15 +44,15 @@ class ChainControllerConfig:
     N_horizon: int = 20
     T_horizon: float = 0.25
     n_mass: int = 5
-    param_interface: ChainAcadosParamInterface = "global"
+    param_splits: ParamSplits = "global"
```
dirkpr: As in `cartpole`, I don't think we want `param_splits` defined through the config. Will probably be the same in the other examples.

leap_c/examples/mass_spring_damper/acados_ocp.py
```diff
@@ -25,87 +27,55 @@ def export_parametric_ocp(
     ocp = AcadosOcp()
     ocp.solver_options.N_horizon = N_horizon
     manager = AcadosParameterManager(N_horizon=N_horizon)
+    spaces: OrderedDict[str, gym.spaces.Box] = OrderedDict()
+
+    def register_learnable(
```
dirkpr: At this level, we only consider `differentiable` -> `register_differentiable`.

leap_c/examples/mass_spring_damper/acados_ocp.py
```diff
@@ -176,13 +146,15 @@ def export_parametric_ocp(
     ocp.solver_options.hessian_approx = "EXACT"
     ocp.solver_options.qp_solver = "PARTIAL_CONDENSING_HPIPM"
 
-    return ocp, manager, gym.spaces.Dict(spaces)
+    param_space = gym.spaces.Dict(spaces)
+
+    return ocp, manager, param_space, manager.default_param_dict(param_space.keys())
 
 
 if __name__ == "__main__":
```
dirkpr, line 156: NOTE: Can delete this after the repo split?

leap_c/examples/pointmass/acados_ocp.py
```diff
@@ -33,85 +32,48 @@ def export_parametric_ocp(
     q_diag_sqrt_val = np.array([1.0, 1.0, 1.0, 1.0])
     r_diag_sqrt_val = np.array([0.1, 0.1])
 
-    splits = "stagewise" if param_interface == "stagewise" else "global"
-    spaces: list[tuple[str, gym.spaces.Box]] = []
+    splits = "stagewise" if param_splits == "stagewise" else "global"
+    spaces: OrderedDict[str, gym.spaces.Box] = OrderedDict()
+
+    def register_learnable(
```
dirkpr, line 37: See previous comment. Would rename.

leap_c/planner.py
```diff
@@ -31,7 +31,7 @@ def forward(
         self,
         obs: Tensor,
         action: Tensor | None = None,
-        params: dict[str, Tensor | ndarray] | None = None,
+        params: Any = None,
```
dirkpr, line 30: Why do we need to change to `Any` here?

dirkpr (changes requested):
One issue about the `param_splits`. The other comments are more cosmetic maybe.

leap_c/examples/cartpole/planner.py
JasperHoffmann: So the question is should we offer an explicit `default_param` or not :) But probably it is helpful to do so.

Regarding your question, yes it could be confusing if you have multiple params and part of them are not updated because someone forgot to set `differentiable = True`. Then most likely there is no error message. However, we could add a type checker for `AcadosDiffMpcTorch`, if an input still has a needs_grad where we can not calcualte a grad we can raise an `error`. What would be even nicer is if we would flexible compile the OCP version requested, but this is too much for now.

leap_c/examples/cartpole/planner.py
JasperHoffmann: Will update :)

leap_c/examples/cartpole/planner.py
JasperHoffmann: From my side we can also make the example not flexible, or how would you do it otherwise. Or should we do two `create_acados_ocp`?

leap_c/planner.py
JasperHoffmann: I think having any typehint here is wrong. I would probably also remove the one for obs :) More flexiblity? Do you agree?

leap_c/examples/cartpole/acados_ocp.py
JasperHoffmann: Hm... I thought default_param is just all the params you want??? Why is it misleading?

leap_c/examples/cartpole/planner.py
dirkpr: Yes, I think not flexible would be the way here. Directly in `export_parametric_ocp` something like:

```
    xref1 = manager.register_parameter(
        "xref1",
        default=np.array([0.0]),
        differentiable=True,
        splits="global",
    )
```

leap_c/planner.py
dirkpr: I agree it is more flexible yes, but also less documenting. Do we need to allow more than `Tensor | dict[str, Tensor | ndarray] | None`? I think this is more a general discussion on type hints

leap_c/examples/cartpole/acados_ocp.py
dirkpr: We are sort of coupling the parameters here with what `ParameterizedController.forward` expects. But at the state of defining the `AcadosOcp`, people could expect `default_param` to cover all parameters that are registered. Can fix this in the documentation later.

dirkpr (approved):

JasperHoffmann:
I think therre is still an incosistency in the current parameter manager with respect to what happens if we add splits by selecting specific split points :) Will check again tomorrow and open an issue.

dirkpr (approved):

Issue #310, opened 2026-06-26
Simplify sensitivity retrieval


JasperHoffmann:
This should be simplified we should not offer this anymore in the abstract interface for now?

_Originally posted by @JasperHoffmann in https://github.com/leap-c/leap-c/pull/302#discussion_r3481314484_

JasperHoffmann:
Use the nice idea of @MazenAmria. We can remove all the hand written parts considering sensitivity retrieval. Should be added into our minimal torch example.

```python
J_y = torch.autograd.functional.jacobian(lambda price: mpc_layer(x0, param={"price": price})[1], price) 
```
This would calculate $\mathrm{d}u_0/\mathrm{d} p_\mathrm{price}$.  Adding `ctx` to the call will actually use caching of previous forward, backward passes.

As an optional caching mechanism we could check whether in the sensitivity call of `DiffAcadosMpc` is just the Jacobian retrieval call by PyTorch. Otherwise there is an potential of multiplying the identity matrix times the already calculated sensitivity. We can for now mark it as an optional todo.

The real nice thing is that there will never be custom gradient splitting again, as we can add any transformation of the params before.

All solved 🌹

JasperHoffmann:
https://github.com/pytorch/pytorch/blob/v2.12.0/torch/autograd/function.py#L617

Maybe we can mark our function as once differentiable :)

PR #317, opened 2026-06-28
Replace manual sensitivity API with PyTorch autograd

JasperHoffmann:
This implements the idea of issue #310.

This PR depends on #316 , as some param interfaces are broken.

Removes the 3-layer sensitivity pass-through (`ParameterizedPlanner.sensitivity` → `AcadosPlanner.sensitivity` → `AcadosDiffMpcTorch.sensitivity`) in favor of standard `torch.autograd.functional.jacobian` over the Diff. MPC layer.

### Why

Sensitivities (e.g. `du0/dp`) were retrieved via a custom 3-layer API returning numpy arrays, which severed the autograd graph. `functional.jacobian` produces the same numbers (verified against acados adjoint + finite differences) and returns a `Tensor` with gradient history — no `torch.from_numpy` needed. Furthermore, as we do a lot of transformation of the parameters it reveals a lot of implementation details. The original sensitvities can be still accessed in AcadosDiffMpcFunction.

### How to retrieve sensitivities now

**Cold (no prior solve):**
```python
du0_dp = torch.autograd.functional.jacobian(
    lambda p: diff_mpc(x0, params={"price": p})[1], price
)
```

**Warmstarted (reuse ctx from a prior forward — faster convergence):**
```python
ctx, *_ = diff_mpc(x0, params={"price": price})
du0_dp = torch.autograd.functional.jacobian(
    lambda p: diff_mpc(x0, params={"price": p}, ctx=ctx)[1], price
)
```

Both do 1 forward solve + 1 adjoint call (cached in ctx); subsequent output elements are einsum contractions on the cached Jacobian.

Note, that currently we warmstart but not resolve entirely on caching as the inputs are changed. @dirkpr Should we update the caching already or should we keep it simple for now. Maybe we can do it in a seperate `PR`???

### Changes

- **Removed** `SensitivityOptions`, `ParameterizedPlanner.sensitivity()`, `ControllerFromPlanner.jacobian_action_param()`, `ParameterizedController.jacobian_action_param()`, `AcadosPlanner.sensitivity()`, `AcadosDiffMpcTorch.sensitivity()`, `TO_ACADOS_DIFFMPC_SENSOPTS`, and `test_planner.py`
- **Refactored** `mpc_actor.py` entropy correction + `sac_fop.py`/`open_chunk_fop.py` logging to use `functional.jacobian`
- **Added** `test_jacobian_matches_sensitivity` (cold/warm/distracting-ops vs acados adjoint, 4 OCP parametrizations)
- **Updated** tutorial + benchmark script (`scripts/cartpole/compare_sensitivity.py`)
- **Kept** `AcadosDiffMpcFunction.sensitivity()` in `diff_mpc.py` — backward pass uses it internally

### Open question: solver-state caching

Each `functional.jacobian` call re-solves the QP (warmstarted, but not skipped). The existing `_PREPARE_BACKWARD_CACHE` uses identity comparison on `ctx`, so a fresh `ctx` per call always re-factorizes. A future optimization could cache the acados capsule (QP matrices + factorization) on `ctx` and skip re-preparation when the solver state is still valid — see the existing TODO in `prepare_solver.py:12`.


tests/leap_c/examples/compare_sensitivity_timings.py
dirkpr, line 1: Good for testing the effects of more efficient batch Jacobian computations or caching as mentioned in the PR description. But we should probably keep the file elsewhere. Will have to be removed anyways when doing the split to a separate example repo.

tests/leap_c/ocp/acados/test_acados_torch.py
```diff
@@ -545,6 +547,98 @@ def test_sensitivity(
...
+    du0_dp_global = torch.as_tensor(diff_mpc.diff_mpc_fun.sensitivity(ctx, "du0_dp_global"))
+    dvalue_dp_global = torch.as_tensor(diff_mpc.diff_mpc_fun.sensitivity(ctx, "dvalue_dp_global"))
+
+    # Extract the slice of p_global corresponding to this param
+    s, e = manager._learnable_parameter_store.indices[param_name]
+    du0_dp_ref = du0_dp_global[:, :, s:e]  # (B, nu, param_dim)
+    dvalue_dp_ref = dvalue_dp_global[:, :, s:e]  # (B, 1, param_dim)
+
+    # Helper: extract per-batch Jacobian from functional.jacobian output
+    def _extract_batch_jac(jac: torch.Tensor) -> torch.Tensor:
```
dirkpr, line 621: Could be useful to have in `sac_fop.py` and `mpc_actor.py`

tests/leap_c/ocp/acados/test_acados_torch.py
```diff
@@ -545,6 +547,98 @@ def test_sensitivity(
         )
 
 
+def test_jacobian_matches_sensitivity(
```
dirkpr, line 550: We should also have this test for stagewise parameters. I can add that.

dirkpr (changes requested):

dirkpr (approved):

Issue #280, opened 2026-03-19
Clean Trainer Classes


JasperHoffmann:
Currently, both for `SacZopTrainer` and `SacFopTrainer` we subclass the cfg from `SacTrainer`. However, this is not ideal, as it introduces additional complexity and  we have currently dead cfg attributes that are not used in `SacZopTrainer` and `SacFopTrainer`. Thus, we should introduce them separately.

PR #318, opened 2026-06-29
fixed config issues

JasperHoffmann:
#280 Cleaned up some config issues regarding the trainers.

There was subclassing of trainer cfgs. This `PR` makes every dataclass of the trainers flat, to make it easier to understand.


dirkpr (approved):
Looks good!

PR #323, opened 2026-07-01
Fix combine_learnable_parameters_torch indexing for custom splits

JasperHoffmann:
## Bug: combine_learnable_parameters_torch uses wrong indexing for custom splits

`AcadosParameterManager.combine_learnable_parameters_torch` mismatches per-segment and per-stage shapes for `int`/`list[int]` splits.

**What works:** `"global"` (1 segment = 1 stage) and `"stagewise"` (N+1 segments = N+1 stages) — in both cases segment count equals stage count, so the bug is invisible.

**What breaks:** Custom splits like `[2, 5]` with N_horizon=5:
- CasADi expression is correct: creates 2 independent symbols.
- `combine` validates `values.shape[1] == splits[-1]+1` (per-STAGE = 6), rejecting the per-SEGMENT shape `(batch, n_segments=2, ...)`.
- If you pass per-stage to satisfy validation, it indexes `val[:, start]` at stage indices 0 and 3 — silently ignoring 4 of the 6 values.

## Changes
- Add `AcadosParameter.overwrite_shape(N_horizon)` and `broadcasted_default(N_horizon)` helper methods.
- Fix `combine_learnable_parameters_torch` to accept per-segment input `(batch, n_segments, ...)` and index by segment index, not stage start index.
- Update validation to expect `n_segments` instead of `splits[-1]+1` or `N_horizon+1`.
- Use `broadcasted_default` for the no-overwrite path.

## Tests
- Helper shape/default tests for all split types (`global`, `stagewise`, `[2, 5]`, `int=3`).
- Combine tests for `[2, 5]` splits, `int` splits, literal `stagewise` — verify distinct values per segment.
- Default-loading tests (no overwrites) for all split types.
- Uncovered terminal stage (`splits=[2, 4]` with N_horizon=5).
- Gradient flow test (`requires_grad` tensors produce non-None gradients).
- All 213 tests pass (baseline 199 + 14 new).

### DIFF
--- a/leap_c/ocp/acados/parameters.py
+++ b/leap_c/ocp/acados/parameters.py
@@ -120,6 +120,19 @@ class AcadosParameter:
                     stacklevel=2,
                 )
 
+    def overwrite_shape(self, N_horizon: int) -> tuple:
+        if self.splits == "global":
+            return self.default.shape
+        _, ends = _define_starts_and_ends(self.splits, N_horizon)
+        return (len(ends), *self.default.shape)
+
+    def broadcasted_default(self, N_horizon: int) -> np.ndarray:
+        if self.splits == "global":
+            return self.default
+        _, ends = _define_starts_and_ends(self.splits, N_horizon)
+        n_segments = len(ends)
+        return np.tile(self.default, (n_segments, *([1] * self.default.ndim)))
+
 
 @dataclass
 class _ParameterStore:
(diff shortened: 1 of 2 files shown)


dirkpr (approved):

JasperHoffmann:
The failing test `test_scaled_beta_anchor[many-stoch]` in `test_bounded_dist.py` is a pre-existing flaky stochastic test unrelated to this PR. It generates random samples from a `ScaledBeta` distribution and asserts all are within bounds — the `stoch` variant can occasionally produce out-of-bounds values due to numerical precision. It passes on Python 3.12 and 3.13, and this PR does not touch `bounded_distributions.py` or its tests. I've re-run the failed job.

PR #331, opened 2026-07-01
Remove set constraint bound method

JasperHoffmann:
This PR removes an outdated method that introduces state in the AcadosDiffMpc

Commits:
- ruff stuff (f7e3a29)
- removed set_constraints_bounds method (fa6c769)


JasperHoffmann:
We still should document better that we can not diff. through constraints and how to set constraints...

JasperHoffmann:
#335 This is already handled here so this PR is oudated.

PR #336, opened 2026-07-01
Update Documentation

dirkpr:
Wire up automatic API documentation from docstrings and type hints, which was previously absent even though autodoc/napoleon were enabled.

- Add sphinx-autoapi + sphinx-autobuild to the docs extra
- Enable autoapi in conf.py (no-import parsing, so docs build without acados compiled); hide private/dunder/imported members, populate intersphinx, fix sys.path; set napoleon_use_ivar to avoid duplicate-object warnings from dataclass attributes
- Add a curated "Core API" landing page + an API Reference toctree section foregrounding the main public classes
- Add `make livehtml` live-reload preview target and document it
- Ignore the transient autoapi-generated tree
- Remove phantom tutorial/parameter_manager/ script references


JasperHoffmann (approved):
Looks good. Worked for me locally. Now to the docs :D

PR #337, opened 2026-07-02
Diff mpc tutorial

dirkpr:
Notebook on using `AcadosDiffMpcTorch` for `forward`/`backward` pass.


notebooks/diff_mpc_tutorial.py
```diff
@@ -0,0 +1,619 @@
...
+    return
+
+
+@app.cell
+def _(diff_mpc, mo, torch):
+    x0 = torch.tensor([[0.5, 0.0]])  # one observation: displaced 0.5 m, at rest
+
+    ctx, u0, x_traj, u_traj, value = diff_mpc(x0=x0)
+
+    mo.md(
```
JasperHoffmann: maybe we should avoid mixing md cells with code cells

notebooks/diff_mpc_tutorial.py
```diff
@@ -0,0 +1,619 @@
...
+    import casadi as ca
+    import matplotlib.pyplot as plt
+    import numpy as np
+    import torch
+    from acados_template import AcadosOcp
+
+    from leap_c.ocp.acados.parameters import AcadosParameterManager
+    from leap_c.ocp.acados.torch import AcadosDiffMpcTorch
+
+    # acados is double precision — use float64 tensors throughout so gradients
```
JasperHoffmann: This is not needed and probably should not be part of a first tutorial

notebooks/diff_mpc_tutorial.py
```diff
@@ -0,0 +1,619 @@
...
+
+    Because the state is 2-D we can solve the MPC from a whole grid of
+    initial states in a single batched call, then plot:
+
+    - the **value function** $V(x_0)$ — the optimal cost-to-go, a bowl
+      centred at the origin, and
+    - the **policy** $\pi(x_0) = F_0^{*}(x_0)$ — the first optimal action,
+      a feedback law that saturates at the $\pm 0.5$ force limit.
+
+    We evaluate on an *inner* grid $[-1.8, 1.8]^2$ to stay away from the
```
JasperHoffmann: Relevant?

notebooks/diff_mpc_tutorial.py
```diff
@@ -0,0 +1,619 @@
...
+    map_fig
+    return
+
+
+@app.cell(hide_code=True)
+def _(mo):
+    mo.md(r"""
+    ## Differentiability — the whole point
+
+    The solver sits inside the autograd graph, so we can differentiate its
```
JasperHoffmann: is part of

notebooks/diff_mpc_tutorial.py
```diff
@@ -0,0 +1,619 @@
+"""A differentiable-MPC tutorial built around ``AcadosDiffMpcTorch``.
+
+This notebook shows how to take a *parametric* optimal control problem (OCP),
```
JasperHoffmann: I think the tutorial is great. Maybe I would still say the complexity is really high. Potential tutorials:
1. intro_diff_mpc: Define a very short simple OCP, just show you can diff through it get gradients on a high level, to show the capabiltiies
2. using_diff_mpc_coming_from_acados: Maybe this one?
3. planner_interface: Reference to examples, how to get planner and so on.
4. simple_learning_example: Show how to do IL with it?

JasperHoffmann (changes requested):

notebooks/diff_mpc_tutorial.py
JasperHoffmann: Actually I am not completely sure anymore, as at least people understand the fancy printing of marimo :D

JasperHoffmann (approved):
Looks good, more detailed notebooks can move over to leap-c-lab in the long term :) I also removed the minimal example from before.

PR #343, opened 2026-07-04
Improve docstring torch layer

JasperHoffmann:
Improved the docstring of the torch layer.

### DIFF
--- a/leap_c/torch.py
+++ b/leap_c/torch.py
@@ -1,11 +1,13 @@
 """Central interface to use acados in PyTorch."""
 
 from pathlib import Path
+from typing import TYPE_CHECKING
 
 import numpy as np
 from acados_template import AcadosOcp
 
 from leap_c.autograd.torch import create_autograd_function
+from leap_c.diff_mpc.data import validate_forward_inputs
 from leap_c.diff_mpc.function import (
     AcadosDiffMpcCtx,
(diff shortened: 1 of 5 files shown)


leap_c/torch.py
```diff
@@ -18,20 +20,35 @@
...
 
     Accepts a plain ``AcadosOcp`` together with a parameter manager. The parameter manager's
     :meth:`~AcadosParameterManager.combine_differentiable_parameters_torch` is called in the forward
     pass to build a flat differentiable tensor from the ``params`` dict.
 
+    Examples:
+        Forward pass with differentiable parameters; gradients flow back through ``params``::
+
+            >>> diff_mpc = AcadosDiffMpcTorch(ocp, manager)
+            >>> ctx, u_star, x, u, value = diff_mpc(x0, params={"cost_weight": w})
```
dirkpr: I would suggest we rename `u_star` to `u0` everywhere. Two reasons:

- inconsistency with the other outputs which are also optimal and would need a _star
- when passing `u0` to treat the mpc as an action-value function, the output would also be `u0`

dirkpr (approved):
Looks good. One comment that we could address here because it also affects the docstrings.

leap_c/torch.py
JasperHoffmann: Done :)

PR #346, opened 2026-07-08
Remove custom_examples from notebooks.

dirkpr:
Remove notebooks that are more complex to only focus on `getting_started`


JasperHoffmann (approved):
Looks good!

PR #338, opened 2026-07-02
Restructure to core-only: diff_mpc, parameters, layer, torch interface

MazenAmria:
**Motivation:** The old repo bundled a full RL stack (SAC, CrossQ, replay buffers, extractors) and multiple example environments alongside the core differentiable MPC code. This created heavy dependency chains (gymnasium, scipy, PyYAML, tensordict) and made the core diff-MPC layer hard to isolate, test, and reuse.

**What survives:** A lean package with one clearly layered hierarchy:

```
leap_c/
├── torch.py               AcadosDiffMpcLayerTorch    (torch.nn.Module, user-facing entry point)
├── parameters/
│   ├── data.py             _AcadosParameter, _ParameterStore
│   └── manager.py             AcadosParameterManager    (parameter registration + combine logic)
├── diff_mpc/
│   ├── function.py         AcadosDiffMpcFunction     (framework-agnostic, numpy-based solver)
│   ├── data.py             AcadosOcpSolverInput + collate helpers
│   ├── initializer.py      solver warm-start strategies
│   └── utils/              solver creation, preparation, retry
├── autograd/
     ├── function.py.     Base interface for wrapping functions into autograd frameworks.
│   └── torch.py            create_autograd_function  (wraps DiffFunction as torch.autograd.Function)
└── utils/
    ├── repr.py                 shared __repr__ formatting
    ├── collate.py                 collate (how to stack acados ctx objects into batches)
    ├── dependencies.py     require_torch(), require_jax()  (lazy import + clear error messages)
    └── parameters.py       ParamSplits, _define_starts_and_ends
```

**Key changes for users:**

- **No auto-installed backend.** Users install `torch` (or later `jax`) manually. `require_torch()` raises a clear error at import time if it's missing. A `torch` optional extra exists for `uv sync --extra torch`.
- **`AcadosDiffMpcLayerTorch`** is the single entry point. No abstract base class, no planners, no controllers. Create one from an `AcadosOcp` + `AcadosParameterManager` and call `layer(x0, params=...)`.
- **Import paths:** `from leap_c.torch import AcadosDiffMpcLayerTorch`, `from leap_c.parameters.base import AcadosParameterManager`.
- **Tests:** 122 tests across `tests/leap_c/diff_mpc/` (conftest+torch+parameters) and `tests/leap_c/autograd/`.
- **CI:** script/integration tests removed. Unit tests run on Python 3.11-3.13.

PR #347, opened 2026-07-09
Add JAX backend support

MazenAmria:
Rewrites the JAX layer that was prototyped on origin/jax-support against the current main package layout (leap_c/{diff_mpc,parameters,torch.py,utils}), rather than reintroducing main's own restructure (#338) from that branch.

- leap_c/autograd/jax.py: wraps DiffFunction as a jax.custom_vjp callable via jax.pure_callback. Builds a fresh custom_vjp + context box per call instead of one shared at layer-construction time, so overlapping forward calls on the same layer (e.g. two rollouts before either is differentiated) can't clobber each other's backward-pass context.
- leap_c/jax.py: AcadosDiffMpcJax, the JAX counterpart to AcadosDiffMpcTorch. Matches torch's ctx-based warmstarting (forward returns/accepts a ctx) and its u0=None "free variable" semantics (a decision made in main's torch layer after the original branch forked) rather than origin's u0=zeros default, which was quietly computing Q-values instead of V-values.
- combine_differentiable_parameters_jax implemented in parameters/manager.py. Also fixes a batch_size=0 coercion bug (`batch_size or 1`) shared by the torch and jax combine methods.
- pyproject.toml: adds a jax-cuda12 extra so CUDA JAX is a declared, lockable dependency instead of a manual post-install `uv pip install` that a later plain `uv sync` would silently revert to CPU.
- Tests mirror a proportionate subset of the torch suite, plus a jax-vs-torch gradient equivalence check (wrt x0 and a differentiable parameter) that catches composition bugs unit tests on either side miss.

### DIFF
--- a/leap_c/autograd/jax.py
+++ b/leap_c/autograd/jax.py
@@ -40,16 +40,36 @@ def create_jax_callable(
     :func:`leap_c.autograd.torch.create_autograd_function`: passing a ``ctx`` from a previous
     call warmstarts the solve, and the returned ``ctx`` can be fed into the next call.
 
-    The impure acados C solver is isolated inside ``jax.pure_callback``, and gradients flow via
-    ``jax.custom_vjp``. Every call builds a *fresh* ``custom_vjp``-wrapped primal closing over a
-    per-call context box: the forward pass's context is threaded to its own matching backward pass
-    via that box (``AcadosDiffMpcCtx`` isn't a valid JAX pytree, so it can't ride as a
-    ``custom_vjp`` residual). Because the box and the VJP rule are rebuilt on every call rather than
-    shared across calls, concurrent-in-time forward calls on the same layer (e.g. two rollouts
-    before either is differentiated) never clobber each other's context; only the fixed cost of
-    re-registering a VJP rule is paid per call, which is negligible next to the acados solve. Since
-    ``pure_callback`` calls run eagerly, ``jax.grad`` works but ``jax.jit`` of the whole layer (or
-    of ``jax.grad`` thereof) is not supported.
+    The acados C solver is wrapped with ``jax.pure_callback`` so it can sit inside a
+    ``jax.custom_vjp``-differentiated function, and gradients flow via that ``custom_vjp``. This
+    function (``solve``) is itself stateless: the caller's ``ctx`` comes in as an argument and a
+    fresh ``ctx`` goes out as part of the return value, with nothing shared across calls at module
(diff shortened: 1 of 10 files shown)


leap_c/autograd/jax.py
```diff
@@ -0,0 +1,136 @@
...
+    nu: int,
+) -> Callable[..., tuple[AcadosDiffMpcCtx, "jax.Array", "jax.Array", "jax.Array", "jax.Array"]]:
+    """Creates a JAX-callable function with a custom VJP wrapping a ``DiffFunction``.
+
+    The returned function accepts ``(x0, u0, p_global, p_stagewise, ctx=None)`` and returns
+    ``(ctx, sol_u0, x, u, sol_value)``, mirroring
+    :func:`leap_c.autograd.torch.create_autograd_function`: passing a ``ctx`` from a previous
+    call warmstarts the solve, and the returned ``ctx`` can be fed into the next call.
+
+    The impure acados C solver is isolated inside ``jax.pure_callback``, and gradients flow via
```
JasperHoffmann: maybe use not impure :)

leap_c/autograd/jax.py
```diff
@@ -0,0 +1,136 @@
...
+def _to_jax(val: np.ndarray | None) -> "jax.Array | None":
+    return None if val is None else jax.numpy.asarray(val)
+
+
+def create_jax_callable(
+    fun: DiffFunction,
+    N_horizon: int,
+    nx: int,
+    nu: int,
+) -> Callable[..., tuple[AcadosDiffMpcCtx, "jax.Array", "jax.Array", "jax.Array", "jax.Array"]]:
```
JasperHoffmann, line 35: Is it required that we implement specifically for the diff mpc function. I was hoping we can just write generacially as in PyTorch :)

JasperHoffmann (changes requested):
Looks good, I am wondering though wether we can simplify the jax autograd registration.

leap_c/autograd/jax.py
MazenAmria: Both comments are related, `pure_callback` requires inputs/outputs of the callback to be an array with a predefined shape and dtype in order to be compiled. PyTorch doesn't do this JIT compiling step, so we do not need to provide N_horizon, nx, and nu ahead of time. Also, the context is not an array type, so the callback can't return the context object, which is the source of the impurity mentioned in the previous comment.

I've refined the docstrings to make that clearer.

